PR #215, opened 2017-11-23
Enhance etpro scripting

isRyven:
This pr should add proper support for etpro scripting commands:
- revamped `create` command and extended supported spawnvars
- cleaned up fakebrush

- [x] Add `delete` scripting command, to delete entities based on spawnvars (port from ETLegacy)


Aciz:
These are good changes that should be done, however can't fix the merge conflicts so need to re-do this.

PR #469, opened 2020-02-08
added acceleration based speed2 color

wolfinim:
Here's the pull request for the acceleration based speed2 color that I talked to Aciz about yesterday. 

The idea simply is that if you enable the setting `etj_speedColorUsesAccel`, `etj_drawSpeed2`'s color will be white when there is no acceleration and it gradually changes to green or red based on how much your speed has increased or decreased respectively.

I haven't done extensive testing with it, I've only jumped a few gammas to make sure it's working.

### DIFF
--- a/src/cgame/etj_speed_drawable.h
+++ b/src/cgame/etj_speed_drawable.h
@@ -2,11 +2,19 @@
 #include "etj_irenderable.h"
 #include "cg_local.h"
 #include <string>
+#include <list>
 
 namespace ETJump
 {
 	class DisplaySpeed : public IRenderable
 	{
+		struct StoredSpeed
+		{
+			int time;
+			float speed;
+		};
+
+		std::list<StoredSpeed> _storedSpeeds;
 		float _maxSpeed{ 0 };
 		vec4_t _color;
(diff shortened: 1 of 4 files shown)


src/cgame/etj_speed_drawable.h
```diff
@@ -2,11 +2,19 @@
...
 {
 	class DisplaySpeed : public IRenderable
 	{
+		struct StoredSpeed
+		{
+			int time;
+			float speed;
+		};
+
+		std::list<StoredSpeed> _storedSpeeds;
```
haapanen: Is there a reason to use list instead of a vector?

haapanen:
Thanks for the first open source PR! :)

src/cgame/etj_speed_drawable.h
wolfinim: Not really, it was just the first data structure that came to mind when I wanted something that is good at adding elements to the end and removing from the beginning and random access was not necessary.

haapanen: boost has circular buffer etc. for this use. Let's go with this for now.

PR #500, opened 2020-07-18
add cvar to always get an overbounce

ovska:
- add cheat protected toggle cvar `etj_forceOb`
- add forced OB handling for DrawOB and OBwatcher
- clean up some checks for combined worldspawn/surfaceparm settings (`nooverbounce` and `nofalldamage`)

### DIFF
--- a/src/cgame/cg_draw.cpp
+++ b/src/cgame/cg_draw.cpp
@@ -3569,6 +3569,10 @@ static qboolean CG_IsOverBounce(float vel, float initHeight,
                                 float finalHeight, float rintv,
                                 float psec, int gravity)
 {
+	if (etj_forceOb.integer != 0) {
+		return qtrue;
+	}
+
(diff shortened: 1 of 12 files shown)


Aciz:
This is not perfect, there are some minor prediction issues here and there, such as loading on angled surfaces and toggling the cvar just before load.

ovska:
> This is not perfect, there are some minor prediction issues here and there, such as loading on angled surfaces and toggling the cvar just before load.

Probably need to do some extra handling to ignore the handler if teleported (or loaded) on last frame

ovska:
There are still some lingering prediction errors when loading (and probably when teleporting too) to irregular surfaces (slopes or near brush edges). I'll need to look a little deeper, not merge ready yet

ovska:
So far haven't gotten this to work with the quirks of SlideMove.

PR #579, opened 2021-02-17
add constantLight support back to movers

ovska:


Aciz:
Seems to work this time without crashes, gj!

Issue #581, opened 2021-02-19
Add option to bypass wait time in func_static pain-activation


ovska:
Currently `wait` key only affects time added on top of the default wait time.

Aciz:
Might cause issues with too rapid activation (mg42 firing for example), need to check that first.

Aciz:
webhook test

Aciz:
Might cause issues with too rapid activation (mg42 firing for example), need to check that first.

Aciz:
closed via #582

PR #582, opened 2021-02-20
add spawnflag to override func_static activation delay

ovska:
#581


Aciz (commented):
Why make this a spawnflag at all? Why not just replace the 500ms hardcoded delay with the `MIN_ACTIVATION_DELAY` constant?

ovska:
> Why make this a spawnflag at all? Why not just replace the 500ms hardcoded delay with the `MIN_ACTIVATION_DELAY` constant?

Backwards compatibility so existing maps work identically as previous etjump versions.

ovska:
Updated to use `FRAMETIME` constant for minimum wait to be consistent with other entities.

PR #584, opened 2021-02-21
Phasing (nonsolid/solid brushes per client)

ovska:
- Repurposes the following contentparams: `donotenterlarge`, `monsterclip`
- Repurposes the following eFlags for players:
  - `EF_FAKE_MODEL`, was never used for player entities
  - `EF_FORCE_END_FRAME` was only used in deprecated singleplayer code
- Adds entity `target_phase` to control phase A and phase B states
- Adds bitfield worldspawn key `phasing` with the following options:
  - `1` Removes phase flags from player on death/team change
  - `2` Removes phase flags from player on load/goto
  - `4` Gibs player if they are stuck inside phase brush for over 250ms
- Adds worldspawn keys `phaseremapXN` there `X` is `a` or `b`, and `N` is a number from 1 to 32, for clientside remapping of shaders depending on phase flags.
  - Example key/value: `phaseremapa1` `textures/mymap/a_nonsolid;textures/mymap/a_solid`
  - More remaps than 32 could probably be made as there doesn't seem to be an engine limit to that

### DIFF
--- a/src/cgame/CMakeLists.txt
+++ b/src/cgame/CMakeLists.txt
@@ -74,6 +74,7 @@ add_library(cgame MODULE
 	"etj_timerun_view.cpp"
 	"etj_trickjump_lines.cpp"
 	"etj_utilities.cpp"
+	"etj_phasing.cpp"
 	"../game/bg_animation.cpp"
 	"../game/bg_animgroup.cpp"
 	"../game/bg_character.cpp"
--- a/src/cgame/cg_local.h
+++ b/src/cgame/cg_local.h
@@ -22,6 +22,7 @@
 #include "cg_public.h"
 #include "../ui/ui_shared.h"
 #include "etj_awaited_command_handler.h"
+#include "etj_phasing.h"
 
 #define MAX_LOCATIONS       256
 #define POWERUP_BLINKS      5
@@ -4008,6 +4009,7 @@ namespace ETJump
 	extern std::shared_ptr<AutoDemoRecorder> autoDemoRecorder;
 	extern std::shared_ptr<EventLoop> eventLoop;
 	extern std::shared_ptr<PlayerEventsHandler> playerEventsHandler;
+	extern std::shared_ptr<PhaseRemapper> phaseRemapper;
 	void addRealLoopingSound(const vec3_t origin, const vec3_t velocity, sfxHandle_t sfx, int range, int volume, int soundTime);
 	void addLoopingSound(const vec3_t origin, const vec3_t velocity, sfxHandle_t sfx, int volume, int soundTime);
 	bool hideMeCheck(int entityNum);
@@ -4015,6 +4017,7 @@ namespace ETJump
 	void setPhaseMask(pmove_t* pm);
 	void onPlayerRespawn(qboolean revived);
 	void runFrameEnd();
+	void phaseRemap(int eFlags);
 	void DrawCGazHUD();
 	void DrawSnapHUD();
 
--- a/src/cgame/cg_local.h
+++ b/src/cgame/cg_local.h
@@ -4012,6 +4012,7 @@ namespace ETJump
 	void addLoopingSound(const vec3_t origin, const vec3_t velocity, sfxHandle_t sfx, int volume, int soundTime);
 	bool hideMeCheck(int entityNum);
 	int checkExtraTrace(int value);
+	void setPhaseMask(pmove_t* pm);
 	void onPlayerRespawn(qboolean revived);
 	void runFrameEnd();
 	void DrawCGazHUD();
--- a/src/cgame/cg_main.cpp
+++ b/src/cgame/cg_main.cpp
@@ -1010,6 +1010,20 @@ namespace ETJump
 		}
 	}
 
+	void setPhaseMask(pmove_t* pm)
+	{
+		// phase brushes always nonsolid for specs
+		if (pm->ps->pm_type == PM_SPECTATOR)
+		{
+			SETBIT(pm->tracemask, CONTENTS_PHASE_A, 0);
+			SETBIT(pm->tracemask, CONTENTS_PHASE_B, 0);
+			return;
+		}
+
+		SETBIT(pm->tracemask, CONTENTS_PHASE_A, pm->ps->eFlags & EF_PHASE_A);
+		SETBIT(pm->tracemask, CONTENTS_PHASE_B, pm->ps->eFlags & EF_PHASE_B);
+	}
+
 	// General purpose etj_hideMe check for cgame events
 	bool hideMeCheck(int entityNum)
 	{
--- a/src/cgame/cg_main.cpp
+++ b/src/cgame/cg_main.cpp
@@ -28,6 +28,7 @@
 #include "etj_event_loop.h"
 #include "etj_autodemo_recorder.h"
 #include "etj_player_events_handler.h"
+#include "etj_phasing.h"
 
 displayContextDef_t cgDC;
 
@@ -112,6 +113,7 @@ namespace ETJump
 	std::shared_ptr<AutoDemoRecorder> autoDemoRecorder;
 	std::shared_ptr<EventLoop> eventLoop;
 	std::shared_ptr<PlayerEventsHandler> playerEventsHandler;
+	std::shared_ptr<PhaseRemapper> phaseRemapper;
 }
 
 /////////////////////////////////////////////////////////////////////////////////////////////////////////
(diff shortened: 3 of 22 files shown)


isRyven:
The idea is good and i see the potential of it, but the implementation is kinda bad, as far as i understand we basically can only affect two types of volumes at all times, and i can't see this can be extended further. This might be enough for room based runs, where you don't really need to run back a lot in other rooms that might be blocked again. Also legit question, what happens with brushes that end up tjunction-ed with the floor, and then you make it non-solid (the floor on the brush would be voided visually)? Should you then make it non-solid from the beginning upon compiling, and then just force the solid state, then this becomes a hardcoded requirement, you should be specific about, but i guess people would find their way around.

isRyven:
Using playerstate to send phase states is very very weird, and not really scalable or extensible btw.

isRyven:
Couple of years ago i was thinking already how to surpass the limits of the surfaceparms we use, as well as, add some sort of conditioning, and one of my ideas was to utilize the brushmodels instead of going with worldspawn and contents, and actually with brushmodels you probably can easily simulate the whole system you have created, and not be limited with two phase types (A, B) only. Burshmodels would be assigned to entities, and act like triggers, so you don't really need any specific surfaceparms to begin with, basically you are unlimited with what you can do, but to make the behavior per client might be a challenging, however, my take would be is to use the entity callback (see how landmine visibility works), where you could assign the field value per client. Too bad i never had time or motivation to try to implement the volumes this way.

isRyven:
Regarding the PhaseRemapper, the code is extremely bad, not sure why you are using the statics there and not using the instance instead even though you initializing one, my take on it would be next: initialize two phaseshader for each phase types ('A' and 'B'), refactor remapper to store remaps in instance instead, move out the spawnvars parsing out of it, as i am pretty sure remapper should not be responsible for this details, and should be doing one thing, here is my refactoring suggestions, haven't tested if anything works or not, but the idea is there:

```c++
// etj_phase_remapper.h
#pragma once
#include <vector>

namespace ETJump
{
	typedef struct RemapPair {
		std::string from;
		std::string to;
	};

	class PhaseRemapper
	{
	public:
		PhaseRemapper();
		~PhaseRemapper();

		void update(int eFlags);
		void add(const std::string &from, const std::string &to);

	private:
		std::vector<RemapPair> _remaps;
		bool _isActive;
	};
}
```

```c++
void update(bool isActiveNew) 
{
	if (!_remaps.size())
	{
		return;
	}

	if (_isActive == isActiveNew) {
		return;
	}

	_isActive = isActiveNew;

	for (auto& shader : _remaps)
	{
		trap_R_RemapShader(shader.from.c_str(), _isActive ? shader.to.c_str() : shader.from.c_str(), "0");
	}
}

```

```c++
// cg_spawn.cpp
static const int MAX_PHASE_REMAPS = 32;
// moved the initialization method here
void CG_ParsePhaseRemaps(char phaseId, PhaseRemapper &phaseRemapper)
{
	char* buffer;
	char key[64];
	const char *keyBase = "phaseremap";
	const char *separator = ";";

	for (int i = 0; i < MAX_PHASE_REMAPS; i++)
	{
		snprintf(key, sizeof(key), "%s%c%i", keyBase, phaseSuffix, i + 1);

		if (!CG_SpawnString(key, "", &buffer))
		{
			break;
		}

		const char *shFrom = strtok(buffer, separator);
		const char *shTo = strtok(NULL, separator);

		if (!shFrom || !shTo)
		{
			// i'd prefer error in the console instead
			CG_Error("Malformed %s value: %s\n", key, buffer);
			continue;
		}

		if (!trap_R_RegisterShader(shFrom))
		{
			CG_Error("Failed to register shader: %s\n", rmp.to.c_str());
			continue;
		}
	}

	phaseRemapper->add(shFrom, shTo);
}
```

```c++
void SP_worldspawn(void)
{
	// ...

	CG_ParsePhaseRemaps('a', &ETJump::phaseRemapperA);
	CG_ParsePhaseRemaps('b', &ETJump::phaseRemapperB);
{
```

```c++
// calling in active frame or such

ETJump::phaseRemapperA->update(cg.snap->ps.eFlags & EF_PHASE_A);
ETJump::phaseRemapperB->update(cg.snap->ps.eFlags & EF_PHASE_B);
```

Should look way more cleaner, `PhaseRemapper` now doesn't care about the count of phases you have over there, and potentially you could add more (although with this system i doubt so).

ovska:
> The idea is good and i see the potential of it, but the implementation is kinda bad, as far as i understand we basically can only affect two types of volumes at all times, and i can't see this can be extended further.
It _cannot_ be extended further due to running out of contentparms.

> Also legit question, what happens with brushes that end up tjunction-ed with the floor, and then you make it non-solid (the floor on the brush would be voided visually)?
Both surfaceparms have `C_TRANSPARENT` flag in compiler so this shouldn't be a problem. It's also on the mapper to make sure his shaders and map work well together.

> Using playerstate to send phase states is very very weird, and not really scalable or extensible btw.
Prediction, demos and spectating work OOTB when using playerstates. If you know a better way to do it feel free. There are also 4 (not counting phasing) entity flags that can still be used, so surfaceparms would run out much earlier anyway.

> with brushmodels you probably can easily simulate the whole system you have created, and not be limited with two phase types (A, B) only
Brushmodels aren't predicted, and I haven't got any idea how you could make all the nuances of pmove work with brushmodels (trace, slidemove, etc.)

ovska:
> Regarding the PhaseRemapper
Don't remember exactly why the statics were there, probably couldn't get the reference to the shared remapper there.
I'll look into splitting the remappers per phase.

isRyven:
> It cannot be extended further due to running out of contentparms.
Exactly the reason the implementation is very limiting, and you will spend the left contents for it, however, i am not here to limit you as a modder (and most importantly as contributor), i am just stating my concerns over the design, i am not going to hold here the last word or something, it's just my review.

> Prediction, demos and spectating work OOTB when using playerstates. If you know a better way to do it feel free. There are also 4 (not counting phasing) entity flags that can still be used, so surfaceparms would run out much earlier anyway.

The reason i implied really is that using playerstate's entity flags for something like *this* (not the most important feature) might be a waste in general, why exactly this feature should occupy the flags, what if i have X feature that i find more important to reuse free bits (as it will be impossible to redo this afterwards we should be more responsible), it's the same idea that applies to contents essentially, however, i don't really know what's the best approach here though, as an alternative, could use commands to toggle the state, similar to how remapping by the server works, or actually, if there is a free field to be reused in ps, i'd just added the anonymous union around the field, and added a new field inside of it, named something like `general_flags` or such (for clarity), where i would put all the unspecific stuff we could use for our needs.

> Brushmodels aren't predicted, and I haven't got any idea how you could make all the nuances of pmove work with brushmodels (trace, slidemove, etc.)

It's not the brushmodels that would be predictable, but the entities that are referencing the brushmodels. Brushmodels themselves are traceable like any other mover or trigger, the only thing is that there is currently brushmodel limit (256), but as far as i remember, the limit is only exists on cgame side and can be surpassed (i never get to the point of implementing it).

ovska:
How the feature works under the hood matters little to me, so any improvements (replacing eFlags without losing the benefits and such) are welcome.

As for the limitation of only two possible phase brushes, I knew from the get go that this was probably the as far as this feature can go. Two independently controlled contents already allows for very complex jumps without having the limitations that using brushes with mapscript though.

> why exactly this feature should occupy the flags, what if i have X feature that i find more important to reuse free bits

This was a concern, though more for the contentparms. etjump currently uses none of the potentially 6 free eFlags for players so I wasn't worried.
Eating up last two of the realistically usable contentparms was something I discussed with Aciz. Personally I think such a cool feature is worth it.
If brushmodels are used and you can change the contents after map load, there is a plethora of contents that aren't used anywhere that could be repurposed (13 I think).

>  i'd just added the anonymous union around the field, and added a new field inside of it

I don't know how to do this, or even what that means :) The only C++ I've written is for this repository.

isRyven:
> I don't know how to do this, or even what that means :) The only C++ I've written is for this repository.

Let's imagine `onFireStar`t in `playerState_t` structure is actually unused or useless:
```c++
typedef struct playerState_s
{
	...
	int onFireStart;
	...
}
```

we can then basically reuse it for our own context, but instead of using the field named `onFireStart`, we want to be able to alias it under the different name, for this case we can use anonymous union:

```c++
typedef struct playerState_s
{
	...
	union {
		int onFireStart;
		unsigned int etjumpFlags;
	}
	...
}

// almost equal to ps.onFireStart & TJF_PHASE_A
if (ps.etjumpFlags & TJF_PHASE_A) 
{
	// ...
}
```

Both fields will use the same address to store the value, just will be named differently, to make the code clean and more obvious, note this doesn't differ at all from using the `onFireStart` directly on technical level. Now i have no memory which fields are actually unused and if there are any at all, but would be nice to actually have something readable and specifically devoted for our needs as that, instead of directly going with entity flags.

PR #585, opened 2021-02-21
Fix weapons potentially set at WP_NONE when using team command

Aciz:
Not defining weapons when using `team` command leaves them at 0 `(WP_NONE)` which causes mismatch in client session data `client->sess.playerWeapon` & `client->sess.playerWeapon2`

### DIFF
--- a/src/game/g_cmds.cpp
+++ b/src/game/g_cmds.cpp
@@ -1814,8 +1814,9 @@ G_GetDefaultWeaponForClass
 Returns default weapons for class
 =================
 */
-int G_GetDefaultWeaponForClass(gentity_t* ent, char* s, int weapon, bool primary, bool updateclient)
+weapon_t G_GetDefaultWeaponForClass(gentity_t* ent, char* s, bool primary)
 {
+	weapon_t weapon;
 	// sessionTeam isn't set when this is called, so compare against team string
 	if (!Q_stricmp(s, "red") || !Q_stricmp(s, "r") || !Q_stricmp(s, "axis"))
 	{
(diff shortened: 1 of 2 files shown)


src/game/g_cmds.cpp
```diff
@@ -1808,6 +1808,58 @@ void G_SetClientWeapons(gentity_t *ent, weapon_t w1, weapon_t w2, qboolean updat
 	}
 }
 
+/*
+=================
+G_GetDefaultWeaponForClass
+Returns default weapons for class
+=================
+*/
+int G_GetDefaultWeaponForClass(gentity_t* ent, char* s, int weapon, bool primary, bool updateclient)
```
ovska, line 1817: This function could just return `weapon_t` so you don't need a static cast on the caller side.

`weapon`-parameter should be removed as well as it's never used, and this function is only called when we know weapon is `0`.

Suggested signature:
```c
weapon_t G_GetDefaultWeaponForClass(gentity_t* ent, char* s, bool primary, bool updateclient)
```

src/game/g_cmds.cpp
```diff
@@ -1808,6 +1808,54 @@ void G_SetClientWeapons(gentity_t *ent, weapon_t w1, weapon_t w2, qboolean updat
...
+			break;
+		case PC_COVERTOPS:
+			weapon = primary ? WP_STEN : WP_SILENCED_COLT;
+			break;
+		default:
+			break;
+		}
+	}
+
+	return weapon;
```
ovska, line 1856:
```c
return WP_NONE;
```

ovska (changes requested):

ovska (approved):
looks gud

PR #590, opened 2021-03-04
initial macos support

jackeri:
Compiles, and maybe even runs if luck is on your side.


jackeri:
image

Aciz:
Very much appreciated, but as you can see, the build system is gonna need some adjustment to support macOS :D and as I mentioned on Discord already, we have no one who owns a Mac, so testing that things work is a bit difficult. I'll leave this open and we can see how feasible macOS support would be.

isRyven:
Thanks for the patch, i'll see if i find time looking into cleaning it up and making to work with azure pipelines, as it does provide necessary host agent to compile the mod on macos. In the meantime, that boost integration broke the other platforms, i will need to check what headers are missing in deboost to make it working with macos (i might have consciously omitted macos related headers when i made the repack).

isRyven:
Thank you very much for your patch again. 
* I did some cleanups as well as refactored the cmake project files (and some other things). 
* Our CI should now be able to build and test macOs binaries.
* Our deployment CI should be able to produce macOS binaries and pack them along with binaries from other platforms to produce the final release package.
* I have only managed to test the `qagame` binary with etlded, but that's as far as i can go.

PR #607, opened 2021-05-16
Fix last jump speed inaccuracy

Aciz:
Events are processed few frames after `commandTime`, which caused last jump speed to be slightly inaccurate because the value was taken few frames after jump was processed in pmove. This caused issue where jumping at the very last few frames on ramps would print the speed after you lost some due to hitting the ramp.

The new ~stat `STAT_JUMP_SPEED`~ pers `PERS_JUMP_SPEED` can be later utilized to generate a jump speed list, which is planned.

Thanks @suburbski for reporting this issue!

@isRyven I want to double check before merging this that adding a new stat is fine, I don't imagine it being a problem since we have already added some there? `MAX_STATS` is 16, there are now 11 fields.

refs #603

### DIFF
--- a/src/cgame/cg_drawCHS.cpp
+++ b/src/cgame/cg_drawCHS.cpp
@@ -430,7 +430,8 @@ static void CG_CHS_PlaneAngleZ(char *buf, int size)
 
 static void CG_CHS_LastJumpSpeed(char* buf, int size)
 {
-	Com_sprintf(buf, size, "%.0f", cg.lastJumpSpeed);
+	playerState_t* ps = CG_CHS_GetPlayerState();
+	Com_sprintf(buf, size, "%d", ps->stats[STAT_JUMP_SPEED]);
 }
 
 typedef struct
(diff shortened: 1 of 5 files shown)


isRyven:
As an alternative, i think you could store the speed in the `pmext` itself, it would be then prediction based by default, won't need to be transmitted over the net, and hence will always be the latest. And you can have direct access to pmext (`cg.pmext`) field within chs.

The stats are fine, but as always, i'd rather ask if there is any value occupying the precious fields with something unimportant (in general speaking).

Aciz:
Problem with using pmext (as far as I understand) is that it's not transmitted over net = spectating will not display the spectated players jump speeds. I initially did this by adding new field to pmext, but noticed that and went with the event approach.

isRyven:
Sorry if that sounded confusing, but I was referring to the importance of the feature itself to occupy the STATS index. But anyways, it all boils down to how important is it to display the value to the spectators, most of whom won't even have the field displayed in the first place, and how much you can curb the perfectionist inside of you :P.

Aciz:
I absolutely do see value in having it visible to spectators, especially considering this stat can be reutilized later to display jump speed meter (DeFRaG CHS 140, display jump speed history) which I have planned to make later, and have that also visible to spectators.

isRyven:
Alright, i cannot come up with another alternative solution based on your requirements. However, seems that STAT fields are cleared upon respawn, unlike PERS ones, but not sure if that has any value for your feature (or future features).

Aciz:
Hmm a good point, that might be an issue. `PERS` seems to have 12 variables atm (max is 16) , but 4 of them are unused, so there's room for 8 more. I will move this there, I would like to have it not reset on respawns, especially for jump history usage.

PR #608, opened 2021-05-19
Add jump speeds list

Aciz:
`etj_drawJumpSpeeds` - display last 10 jump speeds as either vertical two column list, or a horizontal list.

Cvars:
* `etj_drawJumpSpeeds` - toggle on/off (default `0`)
* `etj_jumpSpeedsX/Y` - X/Y offset (default `0/0`)
* `etj_jumpSpeedsColor` - RGBA color (default `1.0 1.0 1.0 1.0`)
* `etj_jumpSpeedsShadow` - toggle shadow (default `1`)
* `etj_jumpSpeedsStyle` - __0__ = vertical, __1__ = horizontal (default `0`)
* `etj_jumpSpeedsShowDiff` - color jumps speeds based on if it was faster/slower than previous jump (default `0`)
* `etj_jumpSpeedsFasterColor` - RGBA color for faster speeds (default `0.0 1.0 0.0 1.0`)
* `etj_jumpSpeedsSlowerColor` - RGBA color for slower speeds (default `1.0 0.0 0.0 1.0`)


ovska:
nit: I would probably call this "jumpSpeedHistory" instead of "jumpSpeeds", not critical though.

src/cgame/etj_jump_speeds.cpp
```diff
@@ -0,0 +1,202 @@
...
+		float offY = 0.0f;
+		float x1 = 6 + etj_jumpSpeedsX.value;
+		float x2 = 6 + 30 + etj_jumpSpeedsX.value;
+		float y1 = 240 + etj_jumpSpeedsY.value;
+		float y2 = 240 + 12 +etj_jumpSpeedsY.value;
+		bool drawShadow = etj_jumpSpeedsShadow.integer;
+		vec4_t color;
+		bool vertical = !etj_jumpSpeedsStyle.integer;
+
+		if (!etj_drawJumpSpeeds.integer)
```
ovska, line 43: Consider putting this to the top of the function? Could also combine all the checks.
```cpp
if (!etj_drawJumpSpeeds.integer ||
    ps->persistant[PERS_TEAM] == TEAM_SPECTATOR ||
    cg.zoomedBinoc ||
    cg.zoomedScope)
{
    return;
}
```

src/cgame/etj_jump_speeds.cpp
```diff
@@ -0,0 +1,202 @@
...
+	void DrawJumpSpeeds()
+	{
+		playerState_t* ps = getValidPlayerState();
+		float offX = 0.0f;
+		float offY = 0.0f;
+		float x1 = 6 + etj_jumpSpeedsX.value;
+		float x2 = 6 + 30 + etj_jumpSpeedsX.value;
+		float y1 = 240 + etj_jumpSpeedsY.value;
+		float y2 = 240 + 12 +etj_jumpSpeedsY.value;
+		bool drawShadow = etj_jumpSpeedsShadow.integer;
```
ovska, line 39: This would result in less repetition later:
```cpp
auto textStyle = etj_jumpSpeedsShadow.integer
    ? ITEM_TEXTSTYLE_SHADOWED
    : ITEM_TEXTSTYLE_NORMAL;
```

src/cgame/etj_jump_speeds.cpp
```diff
@@ -0,0 +1,202 @@
...
+	void ResetJumpSpeeds()
+	{
+		jumpSpeedHistory.clear();
+		lastDeletedSpeed = 0;
+	}
+
+	void AdjustColors(int jumpNum, vec4_t* color)
+	{
+		auto alpha = (*color)[3];
+		vec4_t fasterColor = { 0.0, 1.0, 0.0, alpha };
```
ovska, line 139: Is there a reason there is a cvar for jump speed in general, but not the diff?

src/cgame/etj_jump_speeds.cpp
```diff
@@ -0,0 +1,202 @@
...
+				}
+				// slower than previous jump
+				else if (currentJumpSpeed < lastDeletedSpeed)
+				{
+					Vector4Copy(slowerColor, *color);
+				}
+				// equal speed
+				else
+				{
+					Vector4Set(*color, 1.0, 1.0, 1.0, alpha);
```
ovska, line 163: Doesn't seem like this function respects `etj_jumpSpeedsColor`?

src/cgame/etj_jump_speeds.cpp
```diff
@@ -0,0 +1,202 @@
...
+			else
+			{
+				Vector4Set(*color, 1.0, 1.0, 1.0, alpha);
+			}
+		}
+	}
+
+	static playerState_t* getValidPlayerState()
+	{
+		return (cg.snap->ps.clientNum != cg.clientNum)
```
ovska, line 196: Test in a demo? If it works, add it to a comment, ie. `// spectating or replay`

src/cgame/etj_jump_speeds.h
```diff
@@ -0,0 +1,44 @@
...
+	constexpr int MAX_JUMPS = 10;
+	std::vector<int> jumpSpeedHistory;		// last 10 jump speeds
+	bool jumpSpeedDeleted;
+	bool resetQueued;
+	int lastDeletedSpeed;					// last jump speed that was deleted from history
+	int team;
+
+	void DrawJumpSpeeds();
+	void UpdateJumpSpeeds();
+	void QueueJumpSpeedsReset();
```
ovska, line 40: What's the use case for this vs. `ResetJumpSpeeds`?

src/cgame/etj_jump_speeds.cpp
Aciz: General code style defines variable for functions at the top like this, and I believe compiler will optimize it anyway. For those checks, I don't really like doing stuff like that when it checks for completely different things.

src/cgame/etj_jump_speeds.cpp
Aciz: Good suggestion, will implement.

src/cgame/etj_jump_speeds.cpp
Aciz: Didn't really think it would be necessary, but can implement ofc.

src/cgame/etj_jump_speeds.cpp
Aciz: This is intended, when `etj_jumpSpeedsShowDiff` is enabled, jumps are colored either white (first jump/equal speed), green (faster), or red (slower).

src/cgame/etj_jump_speeds.h
Aciz: Since the drawing function runs every client frame, I made a queue system for resetting, this allows you to see your previous jump speeds after load/death, and the view is only cleared once you jump again. If I would just call a reset instantly, you would not be able to see your jump speeds after load.

ovska (approved):
LGTM

PR #622, opened 2021-06-30
Fix NJD not always saving jumpTime

Aciz:
In cases where you jump from an NJD surface to another NJD surface within regular jump delay time (so jump delay would normally occur), `jumpTime` wasn't updated. This made it possible to perform NJD -> NJD -> non-NJD jump combinations and not have jump delay on non-NJD surface.


ovska:
Most impressive thing about this is how long it took for this simple bug to be found.

ovska (approved):

PR #627, opened 2021-07-16
Fix gcc 11 warnings

suburbski:


src/cgame/cg_atmospheric.cpp
```diff
@@ -356,7 +356,7 @@ static void CG_RainParticleRender(cg_atmosphericParticle_t *particle)
 	vec3_t     forward, right;
 	polyVert_t verts[3];
 	vec2_t     line;
-	float      len, frac, dist;
+	float      len, /*frac,*/ dist;
```
haapanen, line 359: If frac is not used, why not just remove it instead?

src/cgame/cg_character.cpp
```diff
@@ -162,7 +162,7 @@ static void CG_CalcMoveSpeeds(bg_character_t *character)
 	int           i, j, k;
 	float         totalSpeed;
 	int           numSpeed;
-	int           lastLow, low;
+	int           /*lastLow,*/ low;
```
haapanen, line 165: Same here, why not just remove this?

src/cgame/cg_debriefing.cpp
```diff
@@ -1245,7 +1245,7 @@ void CG_Debriefing_ChatEdit_Draw(panel_button_t *button)
 
 void CG_Debriefing_ChatBox_Draw(panel_button_t *button)
 {
-	int    w, h;
+	int    w /*, h*/;
```
haapanen, line 1248: Here as well.

src/cgame/cg_draw.cpp
```diff
@@ -852,7 +852,7 @@ CG_DrawTeamInfo
 */
 static void CG_DrawTeamInfo(void)
 {
-	int w, h;
+	int w /*, h*/;
```
haapanen, line 855: Here as well.

src/cgame/cg_draw.cpp
```diff
@@ -261,7 +261,7 @@ int CG_DrawFieldWidth(int x, int y, int width, int value, int charWidth, int cha
 {
 	char num[16], *ptr;
 	int  l;
-	int  frame;
+	// int  frame;
```
haapanen, line 264: Here as well.

src/cgame/cg_draw.cpp
```diff
@@ -5208,8 +5208,10 @@ static void CG_DrawPlayerStatus(void)
 	if (etj_HUD_weaponIcon.integer)
 	{
 		// Draw ammo
-		// weap = CG_PlayerAmmoValue(&value, &value2, &value3);
 		value = value2 = value3 = 0;
+		// weap = 
+		CG_PlayerAmmoValue(&value, &value2, &value3);
```
haapanen, line 5213: All of these were previously commented out but now uncommented. Is this intentional or am I missing something?

suburbski: I had commented them out accidentally because of unused return value, should be uncommented

PR #658, opened 2021-08-16
Color time difference white for tied timeruns

suburbski:


Aciz:
Is this handled on game side already? Iirc there a string handling code there as well.

Aciz:
Seems like it is, but it's using z as color. 

https://github.com/etjump/etjump/blob/3efd0421d1c1463bd76500af92521d5ba6907f3a/src/game/etj_timerun.cpp#L311

PR #667, opened 2021-08-23
Fix all CI compiler warnings

suburbski:


Aciz:
Lets hope nothing breaks!

PR #696, opened 2021-09-09
Use RelWithDebInfo build type for release pipeline

suburbski:
Allows debugging the etjump dev server, should be reverted before release

### DIFF
--- a/ci/azure-pipelines-release.yml
+++ b/ci/azure-pipelines-release.yml
@@ -23,13 +23,13 @@ jobs:
 
   - script: |
       mkdir $(buildDir) && cd $(buildDir)
-      cmake .. -DCMAKE_BUILD_TYPE=Release -DBUILD_TESTS=OFF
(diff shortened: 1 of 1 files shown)


Aciz:
Need to remember to turn this off for actual releases, but good to have for now.

isRyven:
As far as i know, there isn't much use in that alone, `msvc` doesn't injects debug symbols into the executables / dlls, but rather stores them in `pdb` files, but in this case, `pdbs` are going to be left in azure's host environment. You probably need to ensure pdbs are included along with pk3 or such.

Aciz:
Linux encodes them in .so files though, right? So at least this will help debug stuff on Linux (which is what prompted @suburbski to do this, he keeps crashing on bounce_b2 on dev server, and can't reproduce it locally).

isRyven:
yeah linux should be good, we should still allow symbols for qagame in future, as far as i remember i did set the debug flag for qagame in old cmake project

Aciz:
`pdb` files are now included for debugging on Windows too.

PR #724, opened 2021-09-23
Add softnoclip

### LINKED ISSUE #620: Add "softnoclip" feature
A noclip-style feature where a player can go to free spectate from their current location. You would not be able to go through solids while in this mode, instead the intent is to allow players to easily spectate jumps from their current location, to help out frustration in some jumps where it's not necessarily obvious on what to do or where to go, when a map doesn't allow noclip.

Aciz:
An extension to `noclip` command - whenever `noclip` is not available, clients will (barring other restrictions such as during timeruns) enter `softnoclip` state. In this state, they act as sepctators: they have normal collision with the world, ~can activate triggers and such~. When toggling `noclip` off, players are returned to their original location, similar to if they had saved before starting and loaded back. `etj_noclipScale` will affect movement speed during this state. This state cannot be triggered on its own, it is automatically entered whenever client tries to use noclip while it's unavailable.

closes #620

### DIFF
--- a/src/cgame/etj_cgaz.cpp
+++ b/src/cgame/etj_cgaz.cpp
@@ -400,6 +400,12 @@ namespace ETJump
 			return true;
 		}
 
+		// a bit weird but this will match players who are softnoclipping
+		if (!(ps->persistant[PERS_TEAM] == TEAM_SPECTATOR) && ps->pm_type == PM_SPECTATOR)
+		{
+			return true;
+		}
+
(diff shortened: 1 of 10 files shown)


Aciz:
This still needs A LOT of thought, as spectators can activate a lot of entities, so this allows for a lot of sequence breaking as it is.

Aciz:
Disabled trigger activation and use of `+activate` while softnoclipping, should be enough to avoid any sort of exploits.

Aciz:
Will re-think this in the future and possibly re-implement, this would be a mess to merge.

PR #735, opened 2021-10-02
Fix gcc sanitizer runtime errors

suburbski:
Fixes (common) modsided errors showing up at runtime with the following instrumentation flags:
`-fsanitize=address -fsanitize-address-use-after-scope -fsanitize=pointer-compare -fsanitize=pointer-subtract -fsanitize=leak -fsanitize=undefined`

### DIFF
--- a/src/cgame/cg_playerstate.cpp
+++ b/src/cgame/cg_playerstate.cpp
@@ -32,7 +32,7 @@ void CG_CheckAmmo(void)
 
 	for (i = 0 ; i < WP_NUM_WEAPONS ; i++)
 	{
-		if (!(weapons[0] & (1 << i)))
+		if (!(static_cast<long>(weapons[0]) & (1L << i)))
 		{
 			continue;
 		}
--- a/src/cgame/cg_playerstate.cpp
+++ b/src/cgame/cg_playerstate.cpp
@@ -14,16 +14,12 @@ CG_CheckAmmo
 If the ammo has gone low enough to generate the warning, play a sound
 ==============
 */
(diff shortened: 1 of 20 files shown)


suburbski:
Also added `-fsanitize=float-divide-by-zero -fsanitize=float-cast-overflow`

src/game/q_shared.h
```diff
@@ -1520,8 +1520,8 @@ typedef enum
 
 typedef struct entityState_s
 {
-	int number;                     // entity index
-	entityType_t eType;             // entityType_t
+	int number;             // entity index
+	int eType;              // changed from entityType_t to int to allow ET_EVENTS + eventNum
```
Aciz, line 1524: I don't suppose this causes any problems @isRyven ? You probably know better.

Aciz (commented):
LGTM but could you check this too @isRyven in case I missed something?

src/cgame/cg_playerstate.cpp
```diff
@@ -32,7 +32,7 @@ void CG_CheckAmmo(void)
 
 	for (i = 0 ; i < WP_NUM_WEAPONS ; i++)
 	{
-		if (!(weapons[0] & (1 << i)))
+		if (!(static_cast<long>(weapons[0]) & (1L << i)))
```
suburbski, line 35: This prevents the overflow here but does the cast make any sense when `weapons[0]` is an int when the flags are set?

isRyven:
is this ready?

suburbski:
yes

isRyven:
alright, thank you

PR #750, opened 2022-01-01
drop boost dependency

isRyven:
refs #465


haapanen (approved):
Looks good to me! Read it all very thoroughly

PR #757, opened 2022-01-08
Disallow and cleanup certain cmds for demo replay

suburbski:


Aciz:
Rather than patching individual functions, we could perhaps split the commands into ones that are/are not allowed during demo playback, and handle the command rejection in a single place, similar to how game commands are handled

https://github.com/etjump/etjump/blob/eba2bca56cf452f7c1184498ae8cdbd8bcc3aa22/src/game/g_cmds.cpp#L5083-L5141

https://github.com/etjump/etjump/blob/eba2bca56cf452f7c1184498ae8cdbd8bcc3aa22/src/game/g_cmds.cpp#L5310-L5332

https://github.com/etjump/etjump/blob/eba2bca56cf452f7c1184498ae8cdbd8bcc3aa22/src/game/g_cmds.cpp#L5357-L5372

src/cgame/cg_consolecmds.cpp
```diff
@@ -1556,14 +1530,28 @@ qboolean CG_ConsoleCommand(void)
...
-		if (!Q_stricmp(cmd, commands[i].cmd))
+		if (!Q_stricmp(cmd, anyTimeCommands[i].cmd))
 		{
-			commands[i].function();
+			anyTimeCommands[i].function();
 			return qtrue;
 		}
 	}
+	
+	if (!cg.demoPlayback)
```
Aciz, line 1543: Rather than checking if we're **not** playing demo, check if we **are** and error out?

```cpp
if (cg.demoPlayback)
{
    CG_Printf( <command> not allowed during demo playback);
    return;
}

for (i = 0 ; i < static_cast<int>(sizeof(noDemoCommands) / sizeof(noDemoCommands[0])) ; i++)
{
    ...
}
```

suburbski: We cannot error out early because in functions further down there is separate parsing for other commands not handled by the two new arrays

Aciz:
Let's go with this for now then, and maybe figure out a way to inform the client later.

PR #769, opened 2022-01-12
Set expanded map alpha to 1

suburbski:


haapanen:
Current one (0.7) looks great. Add a configurable option and use 0.7 as default.

PR #793, opened 2022-09-06
Refactor upper right hud

### LINKED ISSUE #621: Make FPS meter update rate configurable
FPS meter refreshes on every frame, making it hard to read sometimes when you're not quite reaching your maxfps value, or when there are very minor FPS drops. Allow players to set an update interval for it to smooth out the view.

### LINKED ISSUE #709: FPS meter ignores frames while scoreboard is up
Visibly dips from max fps down to some lower value the first few frames after scoreboard is gone, which doesn't reflect true fps

suburbski:
Also adds the `etj_FPSMeterUpdateInterval` cvar

Closes #621, closes #709


src/cgame/etj_upper_right_drawable.cpp
```diff
@@ -0,0 +1,277 @@
...
+
+	if (cg_drawSnapshot.integer)
+	{
+		DrawSnapshot(y);
+	}
+}
+
+void ETJump::UpperRight::DrawTimer(float &y) const
+{
+	#define UPPERRIGHT_X (SCREEN_WIDTH - 6)
```
Aciz, line 112: Maybe just define this in the header as const int? Also it's redefined at L167

src/cgame/cg_main.cpp
```diff
@@ -998,6 +1000,8 @@ cvarTable_t cvarTable[] =
 
 	{ &etj_projection, "etj_projection", "0", CVAR_ARCHIVE },
 	{ &etj_saveMsg, "etj_saveMsg", "^7Saved", CVAR_ARCHIVE },
+
+	{ &etj_FPSMeterUpdateInterval, "etj_FPSMeterUpdateInterval", "250", CVAR_ARCHIVE },
```
Aciz, line 1004: I'm not sure about the default value on this tbh, I would make this 8 t match the original default rate at 125FPS.

Aciz (commented):
You can also remove the `etj_time_utilities.h` include at the top of `cg_draw.cpp` now I think, it wasn't there before I added the current round timer.

PR #817, opened 2022-09-15
Remove cgaz/snaphud indicators from scoreboard

Aciz:
Removes `CGaz` column from default scoreboard, `C/V` indicators from `etj_altScoreboard 1-3`. Adds speclock/hideme indicators to scoreboards 1 & 2.


Aciz:
After further consideration, let's not do this for now. We can make this info available via `/players` command in the future, or even have the visible infos customizeable by client. For now, let's leave it as it is and see about this in the future.

PR #856, opened 2022-12-02
Unlock r_clear

Aciz:
`etj_clear` to unlock `r_clear`.


Aciz:
Cvar tooltip should be extended with ETe options once the code for that is actually implemented
image

PR #916, opened 2023-01-06
Add optimized prediction & fix some prediction related errors

Aciz:
Optimized prediction as per Neil “haste” Toronto's unlagged mod, with some tweaks and fixes from ET: Legacy. Enabled by default, this basically has no "visible" difference to clients (it doesn't change at all on how various parts of prediction work) but simply has a performance benefit - `CG_PredictPlayerState` uses ~5x less CPU time per frame (down from ~5.1% to <1%).


Aciz:
There are some issues with this with `pmove_fixed 1` that need to be resolved before this can be merged.

Aciz:
Everything seems okay with this, will do some additional testing later to validate few cases that I haven't yet tested.

PR #928, opened 2023-01-10
Fix and solidify fireteam cmd handling

suburbski:


Aciz (changes requested):
* The command buffer really doesn't need to be 1024 in size
* Instead of nuking the invite/warn/propose/kick by name, you could just use `ClientNumberFromString` (used in private messages) and change the command buffer size to `MAX_NAME_LENGTH`. `ClientNumberFromString` handles duplicate matches for you and doesn't return a valid number in case of multiple matches.
* Use `ClientNum(ent)` to get clientNum and swap the multiple `ent - g_entities` to that

PR #938, opened 2023-01-14
Add etj_footstepVolume to control footstep/landing related sound volume

### LINKED ISSUE #937: Add an option for adjusting footstep volume
Similar to `etj_weaponVolume` for weapon sounds.

Aciz:
Control sounds generated by movememnt events, including foosteps, landing pain grunts and gib sounds. Gibbing technically isn't related to that exclusively but in ETJump context, it's most likely triggered by falling to your death so makes sense to include.

Closes #937


Aciz:
On second thought, not including gibs with this, there are way too many events unrelated to general movement that can cause a gib.

PR #993, opened 2023-08-02
Add new crosshairs

### LINKED ISSUE #946: Sniper scope overlay draws on top of scoreboard
Also draws over chat, might want to change that too.

Aciz:
Adds 7 new crosshairs, drawn as "vector graphics" by using cgame drawing functions rather than actual image files. These are extension to `cg_drawCrosshair`, occupying values **10-16**.

      

In these previews, white color is primary color (`cg_crosshairColor`) and green is secondary color (`cg_crosshairColorAlt`).

Together with these, there are 4 new cvars:
* `etj_crosshairScaleX/Y` - scales crosshairsize on X/Y axis, range **-5 - 5** (this also works for old crosshairs)
  * This does NOT thicken the lines on new crosshairs, for example scaling X value for crosshair 16 will instead move the lines further apart
  * Negative scale values flip the crosshairs, allowing you to for example flip the triangle, V and T crosshairs upside down
* `etj_crosshairThickness` - adjusts the line thickness of the new crosshairs (range **0-5**)
* `etj_crosshairOutline` - toggles drawing the black outline around the crosshair (default on)

Additionally, this PR fixes/changes the following:
* `cg_crosshairHealth` now respects `cg_crosshairAlpha`
* `cg_crosshairSize` and `cg_crosshairX/Y` now accept floating point values
* CGaz/snaphud no longer draw while zooming (binoc/scope) on 3rd person, or for spectators when a client is zoomed with a sniper 
* Fixes #946 - scoreboard and the rest of the HUD elements are no longer drawn behind binoc/sniper reticles
* Jump speeds list is no longer hidden when zooming with binocs/sniper

Known regressions:
* ~~`cg_drawCrosshair` no longer "loops around" - setting the value to an invalid crosshair number (> 16) won't display any crosshair.~~ None

### DIFF
--- a/src/cgame/cg_drawtools.cpp
+++ b/src/cgame/cg_drawtools.cpp
@@ -143,6 +143,10 @@ range_t AnglesToRange(float start, float end, float yaw, float fov) {
   return ret;
 }
 
+void DrawLine(float x1, float y1, float x2, float y2, const vec4_t color) {
+  DrawLine(x1, y1, x2, y2, 1, 1, color);
+}
+
 /*
 ==============
 DrawLine
@@ -217,10 +221,10 @@ static Point TranslatePoint(Point p, float angle, const vec2_t origin) {
 /*
 ==============
 EdgeFunction
-Edge function to determine if a point (x, y) lies inside or outside the
-triangle defined by p1, p2, and p3. The function returns a positive value if
-the point is inside the triangle, a negative value if it's outside, and zero
-if the point lies on the triangle's edge.
+Edge function to determine if a point (x, y) lies inside or outside a
+triangle. The function returns a positive value if the point is inside a
+triangle, a negative value if it's outside, and zero if the point lies on a
+triangle's edge. This should be called for each vertex pair of a triangle.
 ==============
 */
 static float EdgeFunction(const Point &a, const Point &b, float x, float y) {
--- a/src/cgame/cg_drawtools.cpp
+++ b/src/cgame/cg_drawtools.cpp
@@ -145,25 +145,15 @@ range_t AnglesToRange(float start, float end, float yaw, float fov) {
 
 /*
 ==============
-PutPixel
-Used by CGaz 2
-==============
-*/
-// Dzikie
-void PutPixel(float x, float y) {
-  CG_DrawPic(x, y, 1, 1, cgs.media.whiteShader);
-}
-
-/*
-==============
 DrawLine
-Used by CGaz 2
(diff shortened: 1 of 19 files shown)


src/cgame/cg_drawtools.cpp
```diff
@@ -173,22 +163,139 @@ void DrawLine(float x1, float y1, float x2, float y2, const vec4_t color) {
...
+/*
+==============
+EdgeFunction
+Edge function to determine if a point (x, y) lies inside or outside the
+triangle defined by p1, p2, and p3. The function returns a positive value if
+the point is inside the triangle, a negative value if it's outside, and zero
+if the point lies on the triangle's edge.
+==============
+*/
+static float EdgeFunction(const Point &a, const Point &b, float x, float y) {
```
haapanen, line 226: This takes a, b and x, y
Which ones are the p1-3 mentioned in the comment?
Should it be something is "isWithinTriangle"?

src/cgame/etj_cgaz.cpp
```diff
@@ -258,7 +258,7 @@ void CGaz::render() const {
       scx -= SCREEN_OFFSET_X;
     }
     DrawLine(scx, scy, scx + static_cast<float>(cmd.rightmove),
-             scy - static_cast<float>(cmd.forwardmove), CGaz2Colors[1]);
+             scy - static_cast<float>(cmd.forwardmove), 1, 1, CGaz2Colors[1]);
```
haapanen, line 261: It could just have an overload with the w/h and by default use a size of 1?

src/cgame/etj_crosshair.cpp
```diff
@@ -0,0 +1,220 @@
...
+    auto crosshairOutline = crosshair;
+
+    if (outline) {
+      crosshairOutline.t += 1;
+      Vector4Copy(colorBlack, crosshairOutline.color);
+      Vector4Copy(colorBlack, crosshairOutline.colorAlt);
+    }
+
+    switch (cg_drawCrosshair.integer) {
+      case 10: // vertical line
```
haapanen, line 136: Enum instead of a magical number with comment

haapanen (approved):

src/cgame/cg_drawtools.cpp
Aciz: Afaik "edge function" is the correct mathematical term for this calculation, hence the name, but yeah the comment is a bit weird since it mentions the points, which actually come from `FillTriangle`

PR #1016, opened 2023-08-08
Rework g_blockCheatCvars

Aciz:
Rework `g_blockCheatCvars`.

* `m_pitch` is no longer enforced, this never made much sense and is even more useless nowadays due to most mice being capable of separate X/Y DPI
* cvar now works a a bitflag:
  * value **1** will enforce `cl_yawspeed 0` and `cl_freelook 1`
  * value **2** will enforce `com_maxfps 25 .. x .. 125` with `pmove_fixed 0`


Aciz:
The cvar now acts a a bitflag, so it's easier to possibly expand in the future.

PR #1018, opened 2023-08-09
Fix clang warning on std::move + cleanup

Aciz:
refs #995

### DIFF
--- a/src/game/etj_command_parser.cpp
+++ b/src/game/etj_command_parser.cpp
@@ -27,8 +27,8 @@
 
 #include "etj_string_utilities.h"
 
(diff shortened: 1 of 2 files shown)


src/game/etj_command_parser.cpp
```diff
@@ -91,12 +90,12 @@ void ETJump::CommandParser::expectMultipleTokens(
   }
 
   if (_current != end(_args)) {
-    // we peeked the next one and we want the parser to continue from
+    // we peeked the next one, and we want the parser to continue from
```
haapanen, line 93: Literal nitpicking but the comma is not necessary for this to be valid grammar 😅

haapanen (approved):

PR #1019, opened 2023-08-09
Fix records announcing when no times are set + cleanup

Aciz:
Don't print a banner if no records are set - doesn't make sense to announce that player "broke a record" when there's no record that gets broken.

refs #995

### DIFF
--- a/src/game/etj_timerun_v2.cpp
+++ b/src/game/etj_timerun_v2.cpp
@@ -363,7 +363,9 @@ public:
   std::string message;
 };
 
-void ETJump::TimerunV2::addSeason(const Timerun::AddSeasonParams &season) {
+// do not pass by ref! postTask runs asynchronously and passing by
+// ref will lead to undefined behavior
+void ETJump::TimerunV2::addSeason(Timerun::AddSeasonParams season) {
   _sc->postTask(
       [this, season]() {
         try {
@@ -397,7 +399,9 @@ public:
   std::string message;
 };
 
-void ETJump::TimerunV2::editSeason(const Timerun::EditSeasonParams &params) {
+// do not pass by ref! postTask runs asynchronously and passing by
+// ref will lead to undefined behavior
+void ETJump::TimerunV2::editSeason(Timerun::EditSeasonParams params) {
   _sc->postTask(
       [this, params]() {
         try {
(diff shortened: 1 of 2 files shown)


src/game/etj_timerun_v2.cpp
```diff
@@ -964,16 +926,14 @@ void ETJump::TimerunV2::checkRecord(Player *player) {
...
+       checkpoints, clientNum]() {
         /*
          * We want to check the record for all seasons.
          *
          * If player makes a new overall record, it is broadcasted
-         * If player makes a new relevant season record and there's no overall record,
-         * it is broadcasted
-         * If player makes a new season record, it is printed for the player but not
-         * for others
+         * If player makes a new relevant season record and there's no overall
```
haapanen, line 934: Formatter seems to be messing up the comment formatting a bit but it is what it is

src/game/etj_timerun_v2.h
```diff
@@ -95,32 +91,33 @@ class TimerunV2 {
...
   void startTimer(const std::string &runName, int clientNum,
                   const std::string &playerName, int currentTimeMs);
-  void checkpoint(const std::string &runName,
-                  int clientNum, int checkpointIndex, int currentTimeMs);
+  void checkpoint(const std::string &runName, int clientNum,
+                  int checkpointIndex, int currentTimeMs);
   void stopTimer(const std::string &runName, int clientNum, int currentTimeMs);
-  void addSeason(Timerun::AddSeasonParams season);
-  void editSeason(Timerun::EditSeasonParams params);
+  void addSeason(const Timerun::AddSeasonParams &season);
```
haapanen, line 97: 😬 
All of these commands take the object and pass it to a lambda running on another thread. Replacing the actual copy done here with pass by const ref will actually cause the another thread to not use a copy but instead the original ref (which will be deleted in memory). 

**Here's ChatGPTs explanation**

In C++, when you capture a variable by reference in a lambda, the lambda holds a reference to the original variable, not a copy of it. This applies to const references as well. Here's an example:
```c++
const int x = 42;

auto lambda = [&x]() {
    std::cout << x << std::endl; // Prints 42
};

x = 43; // Error, x is const
```
The lambda above captures x by reference, so it refers to the same const int that exists outside of the lambda. Since x is const, you cannot modify it either inside or outside of the lambda.

However, it is essential to be careful with capturing by reference if the lambda might outlive the captured variable. If the captured variable goes out of scope, the reference inside the lambda will become dangling, leading to undefined behavior.

For example:
```c++
std::function<void()> createLambda() {
    const int x = 42;
    return [&x]() { std::cout << x << std::endl; }; // Warning: returning a lambda that captures a local reference
}

auto lambda = createLambda();
lambda(); // Undefined behavior, x is out of scope
```

haapanen (changes requested):
Do not add pass-by-ref automatically without verifying the actual use case within the function/method.

PR #1020, opened 2023-08-10
Add checkpoint cvars to menus, clamp cp count to 1

Aciz:
* Add checkpoint related cvars to menus (buttons aren't exactly centered since I don't wanna use floats here + menus need to be reworked anyway)
* Changed `etj_checkpointsCount` minimum value to **1**, using **0** would be just same as using `etj_drawCheckpoints 1`, and potentially cause confusion when drawing is enabled, but nothing is drawn.

refs #995


haapanen (approved):
Looks great!

PR #1027, opened 2023-08-17
Fix checkpoint timer breaking on vid_restart

Aciz:
Send current checkpoint times alongside timerun start, so client can restore checkpoint timer after `vid_restart`.

refs #995


tests/timerun_shared_tests.cpp
```diff
@@ -5,54 +5,75 @@ using namespace ETJump;
...
-                                       "1,2,3,4,5,6,7,8,9,10,11,12,13,14,-1,-1"};
+  auto args =
+      std::vector<std::string>{"timerun",
+                               "start",
+                               "1",
+                               "2",
+                               "3",
+                               "-1",
+                               "1,2,3,4,5,6,7,8,9,10,11,12,13,14,-1,-1",
+                               "-1,2,3,4,5,6,7,8,9,10,11,12,13,14,-1,-1"};
```
haapanen, line 71: guess it'd make sense to have a valid test case (e.g. first element should be >= 0) but i dunno

haapanen (approved):

PR #1030, opened 2023-08-21
Fix array out of bounds read in fireteam menus

Aciz:
Fireteam menu string indexing was reliant to there being max 5 menu options, as it was using the last pressed menu as the index. Now that there are 6 after adding rules, this no longer works and causes out of bounds reads if pressing 6th option in the menu, then opening the fireteam vsay menu. Now the indexing is still taken from the last pressed menu option, but is clamped to last element of the array to prevent out of bounds access.

Additionally correction to selecting fireteam members, was using old max limit of 6.

refs #1026


Aciz:
String indexing now uses player classes instead, as the strings it requests are class specific. Spectators also no longer get access to Soldiers vsay menu (spectators have soldier as their class)

There's a bunch of stuff with these vsay menus that actually doesn't work, but I will address them in another PR (see #1031 for example)

PR #1036, opened 2023-09-04
Fix checkpoint timer drawing for timeruns without any checkpoints

### LINKED ISSUE #1034: Checkpoint timer draws for timeruns without checkpoints
No easy way to fix this as far as I can see, as client doesn't get a number of checkpoints in a run, only checkpoint times for current run. Can't check for first checkpoint being `-1` as that will be true for timeruns which have checkpoints, but no record is set.

Aciz:
Prevent checkpoint timer from drawing if current timerun has no checkpoints. This does not prevent it from drawing in cases where there's a record without checkpoint data (few maps have added checkpoints already) but that is desired behavior. This only affects runs which have no checkpoint entities attached to them.

Fixes #1034


src/game/etj_timerun_v2.cpp
```diff
@@ -913,6 +914,20 @@ bool ETJump::TimerunV2::isDebugging(int clientNum) {
   return false;
 }
 
+int ETJump::TimerunV2::indexForRunname(const std::string &runName) {
+  int index;
+  std::string currentRun;
+
+  for (index = 0; index < level.timerunNamesCount; index++) {
+    currentRun = sanitize(level.timerunNames[index], true);
+    if (currentRun == sanitize(runName, true)) {
```
haapanen, line 923: I'd probably extract the sanitize(runName, true) to a variable to reduce the unnecessary sanitization
I'd also consider a cache for these but I guess this isn't executed too often so prolly no point in that

haapanen (approved):
lgtm

src/game/etj_timerun_v2.cpp
Aciz: Done, ty.

Caching could maybe be done yea but we're probably talking about nanoseconds in execution time here so probably not critical, maybe later.

PR #1075, opened 2023-09-22
Fix issues in checkRecord

### LINKED ISSUE #1065: Autodemo won't save new PBs if a season is active, unless overall record is broken
With `etj_ad_savePBsOnly 1`, breaking your own PB when a season is active won't save the demo, only breaking overall record will save it.

Aciz:
* Fix banner prints using wrong time for diff print - was using players own previous best, now uses rank 1 time that was beaten
* Fix `etj_ad_savePBOnly 1` not saving demos when a player made a new personal best while a season was active, but that record wasn't a new overall record (fixes #1065)
* Fix banners not displaying if a player has no previous record set on a run, but breaks either overall or seasonal record as their first record time

### DIFF
--- a/src/game/etj_timerun_v2.cpp
+++ b/src/game/etj_timerun_v2.cpp
@@ -981,16 +981,17 @@ public:
 };
 
 void ETJump::TimerunV2::checkRecord(Player *player) {
-  auto clientNum = player->clientNum;
-  auto activeRunName = player->activeRunName;
-  auto userId = player->userId;
-  auto completionTime = player->completionTime;
-  std::map<std::string, std::string> metadata = {{"mod_version", GAME_VERSION}};
-  std::string playerName = player->name;
-  auto checkpoints =
+  const auto clientNum = player->clientNum;
+  const auto activeRunName = player->activeRunName;
+  const auto userId = player->userId;
+  const auto completionTime = player->completionTime;
+  const std::map<std::string, std::string> metadata = {
+      {"mod_version", GAME_VERSION}};
+  const std::string playerName = player->name;
+  const auto checkpoints =
       Container::map(player->checkpointTimes, [](int time) { return time; });
 
-  auto topRecords =
+  const auto topRecords =
       _repository->getTopRecords(_activeSeasonsIds, _currentMap, activeRunName);
 
   _sc->postTask(
--- a/src/game/etj_timerun_v2.cpp
+++ b/src/game/etj_timerun_v2.cpp
@@ -990,9 +990,12 @@ void ETJump::TimerunV2::checkRecord(Player *player) {
   auto checkpoints =
       Container::map(player->checkpointTimes, [](int time) { return time; });
 
+  auto topRecords =
+      _repository->getTopRecords(_activeSeasonsIds, _currentMap, activeRunName);
+
   _sc->postTask(
       [this, activeRunName, userId, completionTime, playerName, metadata,
-       checkpoints, clientNum]() {
+       checkpoints, clientNum, topRecords]() {
         /*
          * We want to check the record for all seasons.
          *
@@ -1002,8 +1005,6 @@ void ETJump::TimerunV2::checkRecord(Player *player) {
          * printed for the player but not for others
          */
 
-        auto topRecords = _repository->getTopRecords(
-            _activeSeasonsIds, _currentMap, activeRunName);
         std::map<int, Timerun::Record> seasonIdToTopRecord{};
         for (const auto &tr : topRecords) {
           seasonIdToTopRecord[tr.seasonId] = tr;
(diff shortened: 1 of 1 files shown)


src/game/etj_timerun_v2.cpp
```diff
@@ -981,18 +981,22 @@ class CheckRecordResult : public ETJump::SynchronizationContext::ResultBase {
...
+  const auto activeRunName = player->activeRunName;
+  const auto userId = player->userId;
+  const auto completionTime = player->completionTime;
+  const std::map<std::string, std::string> metadata = {
+      {"mod_version", GAME_VERSION}};
+  const std::string playerName = player->name;
+  const auto checkpoints =
       Container::map(player->checkpointTimes, [](int time) { return time; });
 
+  const auto topRecords =
```
haapanen, line 994: This runs on main thread, move to inside postTask.

src/game/etj_timerun_v2.cpp
```diff
@@ -1105,23 +1107,30 @@ void ETJump::TimerunV2::checkRecord(Player *player) {
...
         // Overall record
         if (checkRecordResult->isTopRecordPerSeason[defaultSeasonId]) {
-          auto record =
+          const auto record =
               checkRecordResult->newOwnRecordsPerSeason[defaultSeasonId];
 
-          std::string diffString;
-          if (record.previousTime.hasValue()) {
+          // print banner only if there was an existing record
+          if (seasonIdToTopRecord[defaultSeasonId].time) {
```
haapanen, line 1129: doesn't this crash if seasonIdToTopRecord[defaultSeasonId] is not set?

src/game/etj_timerun_v2.cpp
```diff
@@ -1105,23 +1107,30 @@ void ETJump::TimerunV2::checkRecord(Player *player) {
...
             diffString =
                 "^7(" +
-                diffToString(record.record.time, record.previousTime.value()) +
+                diffToString(record.record.time,
+                             seasonIdToTopRecord[defaultSeasonId].time) +
```
haapanen, line 1133: Maybe this should be extracted to a variable as it's used in at least 2 places?

src/game/etj_timerun_v2.cpp
```diff
@@ -1137,19 +1146,22 @@ void ETJump::TimerunV2::checkRecord(Player *player) {
...
                      ->isTopRecordPerSeason[_mostRelevantSeason->id]) {
-          auto record = checkRecordResult
-                            ->newOwnRecordsPerSeason[_mostRelevantSeason->id];
+          const auto record =
+              checkRecordResult
+                  ->newOwnRecordsPerSeason[_mostRelevantSeason->id];
 
-          std::string diffString;
-          if (record.previousTime.hasValue()) {
+          // print banner only if there was an existing seasonal record
```
haapanen, line 1158: Why not print it always?

src/game/etj_timerun_v2.cpp
```diff
@@ -1137,19 +1146,22 @@ void ETJump::TimerunV2::checkRecord(Player *player) {
...
+          if (seasonIdToTopRecord[_mostRelevantSeason->id].time) {
```
haapanen, line 1159: to a variable

haapanen (changes requested):

src/game/etj_timerun_v2.cpp
```diff
@@ -1006,58 +1012,73 @@ void ETJump::TimerunV2::checkRecord(Player *player) {
...
-            seasonIdToNewRecord[seasonId] = std::move(record);
+            playerNewTopRecordForSeason[seasonId] = std::move(record);
           }
         }
 
         auto result = std::make_unique<CheckRecordResult>();
         result->clientNum = clientNum;
 
-        for (const auto &s : seasonIdToIsNewRecord) {
+        // setup new records for any of our old records we beat
```
haapanen, line 1080: I actually have no idea what this comment means :D

src/game/etj_timerun_v2.cpp
```diff
@@ -1092,46 +1121,53 @@ void ETJump::TimerunV2::checkRecord(Player *player) {
           }
         }
 
+        // this seems redundant...? aren't these false by default anyway?
```
haapanen, line 1124: Is it possible that result->isTopRecordPerSeason is not set for a season?

haapanen (approved):

src/game/etj_timerun_v2.cpp
Aciz: Yeah now looking at it it's not great I guess, better comment would be I guess "Go through our old records and update any record that we beat"

src/game/etj_timerun_v2.cpp
Aciz: I don't think so, that would mean that there are no active seasons I think, and since you can't delete default season, that is never the case.

PR #1077, opened 2023-09-22
Support partial matching for loadcheckpoints + allow cp clearing

Aciz:
* Support `/loadcheckpoints <partial runname>` so full run name doesn't need to be typed out
* Allow clearing loaded checkpoints more easily by specifying rank `-1`
* Checkpoint loading will now fail if requested run has no checkpoint times

### DIFF
--- a/src/game/etj_timerun_v2.cpp
+++ b/src/game/etj_timerun_v2.cpp
@@ -673,7 +673,8 @@ void ETJump::TimerunV2::loadCheckpoints(int clientNum, std::string mapName,
   std::string matchedRun;
   std::string errMsg;
   int matchedCount = 0;
-  const auto matchedRuns = _repository->getRunsForName(mapName, runName, false);
+  const auto matchedRuns =
+      _repository->getRunsForName(mapName, runName, false, true);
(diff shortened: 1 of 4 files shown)


Aciz:
Now handles color codes and case sensitivity too.

src/game/etj_timerun_v2.cpp
```diff
@@ -672,19 +675,77 @@ void ETJump::TimerunV2::loadCheckpoints(int clientNum, std::string mapName,
...
       },
-      [this, clientNum, runName, rank](auto r) {
-        auto result = dynamic_cast<LoadCheckpointsResult *>(r.get());
+      [this, clientNum, rank](auto r) {
+        const auto result = dynamic_cast<LoadCheckpointsResult *>(r.get());
+        const auto runName = result->matchedRun;
+
+        // bail out if no checkpoints are present
+        const int noCheckpointSet = TIMERUN_CHECKPOINT_NOT_SET;
+        if (std::all_of(result->checkpoints.begin(), result->checkpoints.end(),
```
haapanen, line 740: replace with if result->checkpoints[0] == -1 :)

haapanen (approved):

PR #1080, opened 2023-09-28
Deprecate target_printname

Aciz:
This deprecates `target_printname` entity and makes the game just use `target_print` code under the hood. `target_print` already supported everything that `target_printname` does, except the name replacement format specifier was different. The new `target_print` code now handles the old `target_printname` format specifier too.

This also fixes a segfault caused by calling `target_printname` with a message that contained two format specifiers in a row (e.g. `message %s %s message`), and automatically resets the print color to white after name gets replaced. The replacement happens right after the name is inserted, so custom color can still be applied normally.

`target_printname` will still continue to function as expected, with the exception that now if someone for some reason specified `[n]` in their message, that will also be replaced with the activators name.

### DIFF
--- a/src/game/etj_string_utilities.cpp
+++ b/src/game/etj_string_utilities.cpp
@@ -269,6 +269,18 @@ void ETJump::StringUtil::replaceAll(std::string &input, const std::string &from,
   }
 }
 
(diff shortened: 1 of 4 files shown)


Aciz:
I want to do a bit more testing with this still so it'll be a draft for now and can be merged later after more testing is done.

PR #1096, opened 2023-11-26
Fix unchecked dynamic_cast results in timerun functions

Aciz:
Check that we get a valid pointer from `dynamic_cast` calls on various timerun functions, and fix missing/incorrect logger format param for `checkRecord` logger call.

### DIFF
--- a/src/game/etj_timerun_v2.cpp
+++ b/src/game/etj_timerun_v2.cpp
@@ -252,6 +252,10 @@ void ETJump::TimerunV2::clientConnect(int clientNum, int userId) {
         auto clientConnectResult =
             dynamic_cast<ClientConnectResult *>(result.get());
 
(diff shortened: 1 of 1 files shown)


src/game/etj_timerun_v2.cpp
```diff
@@ -252,6 +252,10 @@ void ETJump::TimerunV2::clientConnect(int clientNum, int userId) {
         auto clientConnectResult =
             dynamic_cast<ClientConnectResult *>(result.get());
 
+        if (!clientConnectResult) {
+          throw std::runtime_error("clientConnectResult is NULL");
```
haapanen, line 256: Think we could technically just assert this and crash the application as this should never happen but should be fine this way too

haapanen (approved):

Issue #1114, opened 2023-12-08
Improve !seasons command output


Aciz:
* Don't display `Default` season, it's always active and cannot be deleted/renamed
* Separate active/upcoming/past seasons instead of grouping everything together
* Make start/end date listing more legible, it's currently a bit hard to read

PR #1118, opened 2023-12-11
Improve !seasons output + refactor active seasons handling

Aciz:
New format for `!seasons` command output.

image

* Seasons grouped to active/upcoming/past seasons
* Default season is no longer visible
* Start/end dates are now displayed in a more legible format - time is also no longer displayed as it will always be `00:00:00` anyway, and one can reasonably assume that is the case always.

This also refactors the way active seasons are parsed. On init, we now parse all seasons, and group them into active, upcoming and past seasons. `!seasons` command parses all these separately to group the output to active, upcoming and past seasons. `!add/edit/delete-season` commands call this parsing function, so the vectors that hold the seasons per state are updated on the fly, and `!seasons` command will always be up to date.

refs #1114

### DIFF
--- a/src/game/etj_time_utilities.h
+++ b/src/game/etj_time_utilities.h
@@ -24,15 +24,10 @@
 
 #pragma once
 
-#ifdef min
-  #undef min
-#endif
-#ifdef max
-  #undef max
-#endif
 #include <iomanip>
 #include <string>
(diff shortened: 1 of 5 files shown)


haapanen (approved):
Looks great!

PR #1130, opened 2023-12-18
add numPortals to client, add target_portal_relay entity

ovska:
- require spawnflag for activating targets of `target_remove_portals` for backwards compatibility
- fire all targets of `target_remove_portals` instead of random one
- add `client->numPortals` to track portalgun portals shot in since last reset (e.g. team change, timerun start, death, reset by entity)
- add entity `target_portal_relay` that fires targets if `numPortals` is <= `"maxportals"` key

### DIFF
--- a/src/game/g_target.cpp
+++ b/src/game/g_target.cpp
@@ -1618,6 +1618,26 @@ void SP_target_remove_portals(gentity_t *self) {
   }
 }
 
+void target_portal_relay_use(gentity_t *self, gentity_t *other,
+                               gentity_t *activator) {
+  if (!activator ||
+      !activator->client ||
+      activator->client->sess.sessionTeam == TEAM_SPECTATOR ||
+      !self->target) {
+    return;
+  }
+  // don't need to check health, see comment in player_die
+
+  if (activator->client->numPortals <= self->count) {
+    G_UseTargets(self, activator);
+  }
+}
+
+void SP_target_portal_relay(gentity_t *self) {
+  self->use = target_portal_relay_use;
+  G_SpawnInt("maxportals", "-1", &self->count);
+}
+
 void G_ActivateTarget(gentity_t *self, gentity_t *activator) {
   gentity_t *ent = NULL;
   ent = G_PickTarget(self->target);
--- a/src/game/g_target.cpp
+++ b/src/game/g_target.cpp
@@ -1585,10 +1585,7 @@ void target_remove_portals_use(gentity_t *self, gentity_t *other,
 
   if (hadActivePortals) {
     if (self->spawnflags & SF_REMOVE_PORTALS_ACTIVATE_TARGETS) {
-      gentity_t *ent = G_PickTarget(self->target);
-      if (ent && ent->use) {
-          G_UseEntity(ent, self, activator);
-      }
+      G_UseTargets(self, activator);
     }
 
     // play sound to client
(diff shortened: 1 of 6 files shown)


src/game/g_target.cpp
```diff
@@ -1618,6 +1618,26 @@ void SP_target_remove_portals(gentity_t *self) {
...
 
+void target_portal_relay_use(gentity_t *self, gentity_t *other,
+                               gentity_t *activator) {
+  if (!activator ||
+      !activator->client ||
+      activator->client->sess.sessionTeam == TEAM_SPECTATOR ||
+      !self->target) {
+    return;
+  }
+  // don't need to check health, see comment in player_die
```
Aciz, line 1629: I have no idea what this comment means even after reading the comment in `player_die`, what health? Dead players don't activate entities anyway so my mind instantly goes to `self->health` but that doesn't make any sense.

src/game/g_target.cpp
```diff
@@ -1568,43 +1569,42 @@ void target_remove_portals_use(gentity_t *self, gentity_t *other,
...
+    if (!(self->spawnflags & SF_REMOVE_PORTALS_NO_TEXT)) {
+      Printer::SendCenterMessage(ClientNum(activator), "^7Your portal gun portals have been reset.");
+    }
   }
 
-  if (!(self->spawnflags & SF_REMOVE_PORTALS_NO_TEXT)) {
-    Printer::SendCenterMessage(ClientNum(activator), "^7Your portal gun portals have been reset.");
-  }
+  // reset numPortals after everything else so the targeted entity can potentially use it
+  activator->client->numPortals = 0;
```
Aciz, line 1607: This allows for cheesing stuff with save/load (save next to `target_portal_relay`, backup to start and activate reset, load back), but tbh I can't think of a way to fix it in code, just need to be aware of it while mapping I guess.

src/game/g_target.cpp
ovska: Various delayed activations like a button or target_delay could be activated with `numPortals = 0` if you die (ungibbed) before the activation. I can clarify this in the comment, but if there's a chance that you will be revived, despite losing your portals I think the portal count shouldn't be reset.

ovska: e.g. Shoot 2 portals, press button, hit a hurt-brush, and when the event activates your numportals would be 0.

src/game/g_target.cpp
ovska: Wouldn't the cheese work with portals reseting at the start of the activation as well?  If not, I'll fix this since its unlikely the targeted entities would need a portal count.

Aciz (approved):

src/game/g_target.cpp
```diff
@@ -1568,43 +1569,42 @@ void target_remove_portals_use(gentity_t *self, gentity_t *other,
...
+          EV_GENERAL_CLIENT_SOUND_VOLUME,
+          self->noise_index);
 
-  // play sound to client
-  if (self->noise_index) {
-    gentity_t *noiseEnt = ETJump::soundEvent(
-        activator->r.currentOrigin,
-        EV_GENERAL_CLIENT_SOUND_VOLUME,
-        self->noise_index);
+      noiseEnt->s.onFireStart = 255;
```
Aciz, line 1598: This should probably have a `volume` key associated with it? Since you're sending a volume controlled event anyway.

src/game/g_target.cpp
ovska: reseted portals in player_die anyways

PR #1137, opened 2023-12-21
allow having multiple CGAZ huds visible simultaneously

ovska:


src/cgame/etj_cgaz.cpp
```diff
@@ -206,25 +206,23 @@ bool CGaz::beforeRender() {
     wishspeed = static_cast<float>(ps->speed) * ps->sprintSpeedScale;
   }
 
-  switch (etj_drawCGaz.integer) {
-    case 1:
-      UpdateDraw(wishspeed, pm->pmext->accel);
+  if (etj_drawCGaz.integer) {
```
Aciz, line 209: This is redundant, cvar is already checked to be non-zero in `canSkipDraw()`.

PR #1152, opened 2024-01-03
Add custom vote editing via admin commands

Aciz:
Adds admin commands to edit customvotes on the server.

* `!add-customvote` - adds a new custom vote list
  * Parameters (all required):
    * `--name [name]` - name of the list
    * `--full-name [callvote_text]` - full name displayed in callvote
    * `--maps [map list]` - maps to include in the list, space delimited
  * If no custom vote file is present, will create a new one
* `!delete-customvote` - deletes a custom vote list
  * Parameters (all required)
    * `--name [name]` - name of the list to delete
* `!edit-customvote` - edits an existing custom vote list
  * Parameters:
    * `--list [name]` - custom vote list to edit (required)
    * `--name [newname]` - sets new name for the list (optional)
    * `--full-name [callvote_text]` - sets new full name for the list (optional)
    * `--add-maps [maps to add]` - adds maps to the list, space delimited (optional)
    * `--remove-maps [maps to remove]` -  removes maps from the list, space delimited (optional)

Command access is provided with admin flag `c`.

Additional changes:
* `customvotes` alias for `listinfo` command
* Add missing autocomplete for `generateCustomvotes` and `generateMotd`
* New command `readCustomvotes` to reload custom vote file

### DIFF
--- a/src/game/etj_commands.cpp
+++ b/src/game/etj_commands.cpp
@@ -26,7 +26,6 @@
 #include "etj_commands.h"
 
 #include "etj_command_parser.h"
-#include "etj_command_parser.h"
 #include "etj_local.h"
 #include "etj_save_system.h"
 #include "etj_session.h"
@@ -82,15 +81,13 @@ const char TOKENS = 'V';
 const char SETLEVEL = 's';
 const char MOVERSCALE = 'v';
 const char TIMERUN_MANAGEMENT = 'T';
+const char CUSTOMVOTES = 'c';
 } // namespace CommandFlags
 
-// refactor admin system to use the new command parser instead of
-// continuing to use this.
-// This is only in use to keep the scope of the timerun improvement
-// reasonable
-auto deprecated_getCommand(const std::string &commandPrefix, int clientNum,
-                           const ETJump::CommandParser::CommandDefinition &def,
-                           Arguments args) {
+namespace ETJump {
+opt<CommandParser::Command>
+getOptCommand(const std::string &commandPrefix, int clientNum,
+              const CommandParser::CommandDefinition &def, Arguments args) {
   auto command = ETJump::CommandParser(def, *args).parse();
 
   if (command.helpRequested) {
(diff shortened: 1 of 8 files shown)


src/game/etj_commands.cpp
```diff
@@ -2222,6 +2224,112 @@ bool RockTheVote(gentity_t *ent, Arguments argv) {
...
+  const auto optCommand =
+      ETJump::getOptCommand("add-customvote", clientNum, def, argv);
+
+  if (!optCommand.hasValue()) {
+    return false;
+  }
+
+  auto command = optCommand.value();
+
+  const std::string &name = command.options["name"].text;
```
haapanen, line 2251: can these fail?

src/game/etj_commands.cpp
```diff
@@ -2222,6 +2224,112 @@ bool RockTheVote(gentity_t *ent, Arguments argv) {
...
+
+  const auto optCommand =
+      ETJump::getOptCommand("delete-customvote", clientNum, def, argv);
+
+  if (!optCommand.hasValue()) {
+    return false;
+  }
+
+  auto command = optCommand.value();
+  const std::string &name = command.options["name"].text;
```
haapanen, line 2277: can this fail

src/game/etj_custom_map_votes.cpp
```diff
@@ -243,3 +446,34 @@ bool CustomMapVotes::isValidMap(const std::string &mapName) {
   return G_MapExists(mapName.c_str()) && mapName != level.rawmapname &&
          !MapStatistics::isBlockedMap(mapName);
 }
+
+bool CustomMapVotes::writeCustomvoteFile(const std::string &file,
```
haapanen, line 450: I'd probably actually just implement a simple generic JSON interface e.g. 
JSON::readFile(file) and JSON::writeFile(file, root) 
This was probably done this way previously but above would make more sense as there's nothing specific to custom map votes

haapanen (approved):
Maybe refactor JSON writing to a separate JSON class that just handles that? Looks good to me though

src/game/etj_custom_map_votes.cpp
Aciz: Yeah this would actually make more sense, will implement some basic thing that we can expand later.

Aciz: Added a very simple `JsonUtils` class that simply has these read/write funcs in it, can be expanded later when needed. Should probably refactor the rest of our json stuff into this too but that can be done later.

src/game/etj_commands.cpp
Aciz: No, `optCommand` has no value if name is empty so the if check above fails.

src/game/etj_custom_map_votes.cpp
haapanen: Nice

PR #1175, opened 2024-01-10
etj_CGazDrawSnapZone

ovska:
cvar to highlight current snapzone in cgaz1 instead of the small accel zone (light green by default)

image

### DIFF
--- a/src/cgame/etj_cgaz.cpp
+++ b/src/cgame/etj_cgaz.cpp
@@ -26,6 +26,7 @@
 // https://github.com/Jelvan1/cgame_proxymod/
 
 #include "etj_cgaz.h"
+#include "etj_snaphud.h"
 #include "etj_utilities.h"
 #include "etj_pmove_utils.h"
 #include "../game/etj_numeric_utilities.h"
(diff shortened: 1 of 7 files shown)


src/cgame/etj_cgaz.cpp
```diff
@@ -115,12 +116,44 @@ void CGaz::UpdateDraw(float wishspeed, float accel) {
...
   drawVel = atan2f(pm->pmext->velocity[1], pm->pmext->velocity[0]);
 }
 
+float CGaz::UpdateDrawSnap() {
+  // don't highligh snapzone on very low velocities
+  if (!etj_CGazDrawSnapZone.integer || state.vf < state.wishspeed) {
+    return NAN;
+  }
+
+  Snaphud::CurrentSnap cs = Snaphud::getCurrentSnap(*ps, pm);
```
Aciz, line 131: This can be a const.

src/cgame/etj_cgaz.cpp
```diff
@@ -232,20 +258,30 @@ void CGaz::render() const {
       fov = Numeric::clamp(etj_CGazFov.value, 1, 179);
     }
 
+    const auto zone1 = drawMin;
+
+    const auto zone2 = std::isnan(drawSnap) ? drawOpt
```
Aciz, line 263: Please let's not do this :D separate this into if/elses.

src/cgame/etj_snaphud.cpp
```diff
@@ -417,14 +420,32 @@ bool Snaphud::inMainAccelZone(const playerState_t &ps, pmove_t *pm) {
...
 
+  cs.opt = opt;
+  cs.yaw = yaw;
+  cs.snap = snap;
+  cs.rightStrafe = rightStrafe;
+  return cs;
+}
+
+bool Snaphud::inMainAccelZone(const playerState_t &ps, pmove_t *pm) {
+  auto cs = getCurrentSnap(ps, pm);
```
Aciz, line 435: This can be a const too.

Aciz (changes requested):
Missing menu entry but that's fine, I can take care of that. However this does not work with non-default snaphud or cgaz fov, having either at non-0 values breaks the drawing.

Aciz:
Correction: this only works if CGaz and Snaphud use the same fov, if they use different fov, the drawing breaks.

Aciz (approved):

PR #1183, opened 2024-01-11
Accelerometer+color coding

numbfingies:
Added acceleration values instead for etj_drawSpeed2, option "10".
Added color coding for speedometer based on current player accel values and optimal accel values: etj_speedColorUsesAccel "2".
Split some cgaz functions to calculate frame acceleration separately.

### DIFF
--- a/src/cgame/cg_main.cpp
+++ b/src/cgame/cg_main.cpp
@@ -347,6 +347,13 @@ vmCvar_t etj_maxSpeedDuration;
 vmCvar_t etj_speedColorUsesAccel;
 vmCvar_t etj_speedAlign;
 
+vmCvar_t etj_drawAccel;
+vmCvar_t etj_accelX;
+vmCvar_t etj_accelY;
+vmCvar_t etj_accelSize;
+vmCvar_t etj_accelShadow;
+vmCvar_t etj_accelAlign;
+
 vmCvar_t etj_popupTime;
 vmCvar_t etj_popupStayTime;
 vmCvar_t etj_popupFadeTime;
@@ -885,6 +892,12 @@ cvarTable_t cvarTable[] = {
     {&etj_maxSpeedDuration, "etj_maxSpeedDuration", "2000", CVAR_ARCHIVE},
     {&etj_speedColorUsesAccel, "etj_speedColorUsesAccel", "0", CVAR_ARCHIVE},
     {&etj_speedAlign, "etj_speedAlign", "0", CVAR_ARCHIVE},
+
+    {&etj_drawAccel, "etj_drawAccel", "1", CVAR_ARCHIVE},
+    {&etj_accelX, "etj_accelX", "320", CVAR_ARCHIVE},
+    {&etj_accelY, "etj_accelY", "350", CVAR_ARCHIVE},
+    {&etj_accelSize, "etj_accelSize", "3", CVAR_ARCHIVE},
+    {&etj_accelAlign, "etj_accelAlign", "0", CVAR_ARCHIVE},
 
     {&etj_popupTime, "etj_popupTime", "1000", CVAR_ARCHIVE},
     {&etj_popupStayTime, "etj_popupStayTime", "2000", CVAR_ARCHIVE},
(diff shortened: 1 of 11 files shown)


numbfingies:
fixed some logic with coloring for accel/speed

numbfingies:
fixet warnings:)

numbfingies:
While this looks correct, it might be inaccurate or even wrong in snap zone transitions, as we compare current state with predicted/optimal next state.

src/cgame/etj_snaphud.cpp
```diff
@@ -445,6 +445,111 @@ bool Snaphud::inMainAccelZone(const playerState_t &ps, pmove_t *pm) {
...
+      return true;
+    }
+    if (snap > 90 && yaw < std::fmod(snap, 90)) {
+      return true;
+    }
+  }
+
+  return false;
+}
+
```
Aciz, line 552: This entire function is unused..?

Aciz (changes requested):
Needs some cleanup.

The feature is good, however I don't think we should add this as an option for `etj_drawSpeed2`, as you can't then draw accel at the same time as regular speed. I think the better solution would be to make this into it's own cvar (`etj_drawAccel` + adjustment stuff).

numbfingies:
pls review <3

src/cgame/etj_speed_drawable.cpp
```diff
@@ -90,43 +130,185 @@ void ETJump::DisplaySpeed::resetMaxSpeed() {
...
+  int accelStyle =
+      _shouldDrawAccelShadow ? ITEM_TEXTSTYLE_SHADOWED : ITEM_TEXTSTYLE_NORMAL;
+
+  vec4_t color = {1, 1, 1, 1};
+
+  playerState_t *ps = &cg.predictedPlayerState;
+  const float xyVelocity = VectorLength2(ps->velocity);
+
+  if (etj_speedColorUsesAccel.integer == 1 ||
+      (etj_speedColorUsesAccel.integer == 2 && xyVelocity <= 452 &&
```
Aciz, line 310: Same here, looks like a bad copy/paste.

src/cgame/etj_speed_drawable.cpp
```diff
@@ -139,19 +321,30 @@ void ETJump::DisplaySpeed::render() const {
...
-  CG_Text_Paint_Ext(x - w, y, size, size, color, status.c_str(), 0, 0, style,
+  CG_Text_Paint_Ext(speedX - speedW, speedY, speedSize, speedSize, color,
+                    status.c_str(), 0, 0, speedStyle,
                     &cgs.media.limboFont1);
+
+  if (etj_drawAccel.integer) {
+    CG_Text_Paint_Ext(accelX - accelW, accelY, accelSize, accelSize, color,
+                      accel_string.c_str(), 0, 0, accelStyle,
+                      &cgs.media.limboFont1);
+  }
```
Aciz, line 338: `canSkipDraw` will early out if `etj_drawSpeed2` is 0, so this might still not draw. Really this has no business being inside this speed drawable but I can take care of refactoring that later, this is fine for now as long as the drawing is fixed with `etj_drawSpeed2 0`.

src/cgame/cg_main.cpp
```diff
@@ -886,6 +893,12 @@ cvarTable_t cvarTable[] = {
     {&etj_speedColorUsesAccel, "etj_speedColorUsesAccel", "0", CVAR_ARCHIVE},
     {&etj_speedAlign, "etj_speedAlign", "0", CVAR_ARCHIVE},
 
+    {&etj_drawAccel, "etj_drawAccel", "1", CVAR_ARCHIVE},
```
Aciz, line 896: Let's not make this 1 by default.

src/cgame/cg_main.cpp
```diff
@@ -886,6 +893,12 @@ cvarTable_t cvarTable[] = {
...
+    {&etj_accelX, "etj_accelX", "320", CVAR_ARCHIVE},
+    {&etj_accelY, "etj_accelY", "350", CVAR_ARCHIVE},
+    {&etj_accelSize, "etj_accelSize", "3", CVAR_ARCHIVE},
+    {&etj_accelAlign, "etj_accelAlign", "0", CVAR_ARCHIVE},
```
Aciz, line 900: This isn't implemented at all?

Aciz (changes requested):
It's better now, just few minor things to correct, I can refactor it later to it's own module and add menu setting entries :+1:

src/cgame/cg_main.cpp
numbfingies: whoopsie, copy-pasted and forgot to replace

numbfingies:
updated the skipDrawCheck condition

Aciz (approved):

PR #1185, opened 2024-01-11
Improve entity health bar visuals

Aciz:
Setup taken from ET: Legacy with some tweaks.

image
image


Aciz:
Changed the `enum class` to an `enum` so casting isn't required everywhere.

PR #1187, opened 2024-01-14
Fetch content for deps

suburbski:


Aciz:
This would require internet access to compile a clean build, and repos to actually be available, which is not necessarily the greatest idea.

PR #1191, opened 2024-02-02
Increase UI mempool size

Aciz:
Was getting quite full (86 percent or so) and somehow 2.60b linux wasn't running at all and crashed with UI_Alloc out of memory error.

### DIFF
--- a/src/ui/ui_shared.cpp
+++ b/src/ui/ui_shared.cpp
@@ -73,9 +73,9 @@ static qboolean Menu_OverActiveItem(menuDef_t *menu, float x, float y);
   #endif
 #else
   #if id386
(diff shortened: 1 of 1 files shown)


src/ui/ui_shared.cpp
```diff
@@ -73,9 +73,9 @@ static qboolean Menu_OverActiveItem(menuDef_t *menu, float x, float y);
   #endif
 #else
   #if id386
-    #define MEM_POOL_SIZE (2048 * 1024) // Arnout: was 1024
+    #define MEM_POOL_SIZE (4096 * 1024) // Arnout: was 1024
```
ovska, line 76: Just missing a comment `// aciz: was 2048`

ovska (approved):

PR #1200, opened 2024-02-05
Fix chat mentions breaking chat coloring

### LINKED ISSUE #1193: Chat mentions break chat coloring on team/ft chat
Text is forced to green after a mention regardless of chat mode.

Aciz:
All mentions transformed the rest of the chat to global chat color, regardless if the chat was global, team or fireteam chat.

Fixes #1193

### DIFF
--- a/src/game/g_utilities.cpp
+++ b/src/game/g_utilities.cpp
@@ -619,7 +619,7 @@ std::vector<int> getMatchingIds(const std::string &name) {
   return pidsVector;
 }
 
-std::string interpolateNametags(std::string input) {
+std::string interpolateNametags(std::string input, const int color) {
   std::string interpolated;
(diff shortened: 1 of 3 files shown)


src/game/g_utilities.cpp
```diff
@@ -619,25 +619,30 @@ std::vector<int> getMatchingIds(const std::string &name) {
...
   if (split.size() == 1 || split.size() == 2) {
     return input;
   }
 
-  auto i = 1;
-  auto len = 0;
-  for (len = split.size(); i < len; i += 2) {
+  int i = 1;
+  int len;
+  std::string colorCode = "^";
```
ovska, line 632: nit: Could be a oneliner: `std::string colorCode = "^" + static_cast<char>(color);`

PR #1203, opened 2024-02-05
Fix speaker editor + improve visuals

### LINKED ISSUE #1189: Speaker editor is broken
Likely as a result of #833 

Valgrind throws a bunch of errors about uninitialized values when interacting with the editor, and the edit fields are not filled correctly and don't work.

image
image

Aciz:
`button->text` being `std::string` required some tweaks to the way the strings are handled. UI is also now widescreen corrected.

fixes #1189 
refs #833

### DIFF
--- a/src/cgame/cg_sound.cpp
+++ b/src/cgame/cg_sound.cpp
@@ -683,33 +683,6 @@ static void CG_RenderScriptSpeakers(void) {
   closest = -1;
   minDist = Square(8.f);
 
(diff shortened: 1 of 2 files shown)


Aciz:
Will do a bit more testing a cleanup later.

ovska (approved):

Aciz:
Improved visuals too, so they now match other menu UI elements.

image
image

PR #1210, opened 2024-02-07
Add basic demo compatibility checker + fix client-only coronas on old demos

Aciz:
Adds a basic demo playback compatibility checker. Currently supports parsing only `MAJOR.MINOR.PATCH` version, so some old versions we had (e.g. `-RC3`) will not be differentiated from the stable release. Also does not support git versions.

### DIFF
--- a/src/cgame/cg_draw.cpp
+++ b/src/cgame/cg_draw.cpp
@@ -10,6 +10,7 @@
 #include "../game/etj_numeric_utilities.h"
 #include "../game/etj_string_utilities.h"
 #include "etj_client_rtv_handler.h"
+#include "etj_demo_compatibility.h"
 
 #define STATUSBARHEIGHT 452
 char *BindingFromName(const char *cvar);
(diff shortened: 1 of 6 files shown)


src/cgame/cg_draw.cpp
```diff
@@ -4633,6 +4634,10 @@ void CG_DrawMiscGamemodels(void) {
 }
 
 void CG_DrawCoronas() {
+  if (cg.demoPlayback && !ETJump::demoCompatibility->isCompatible("3.1.2")) {
```
ovska, line 4637: This method should probably accept a hardcoded struct instead so the parsing cost isn't paid on every invocation, e.g.
`isCompatible({3, 1, 2})`

ovska (changes requested):

src/cgame/cg_draw.cpp
Aciz: Assigned the results to static bool instead, I'd rather keep the input as a string for simplicity so that the compatibility checker is responsible for building the version struct.

Aciz: Actually, went with passing the struct rather than string.

ovska (approved):

ovska (approved):

PR #1223, opened 2024-02-10
Fix goto/call/iwant stance exploit

### LINKED ISSUE #1222: Goto allows bypassing stance guards present in /load
When using `goto` or `call` when proned, stance of the other player isn't considered and you can wallstuck your head (load standing into a lower stance) or legs (load proned while facing away from a wall).

Aciz:
These commands bypassed typical save/load stance restrictions, which allowed for wallbugging.

fixes #1222

### DIFF
--- a/src/game/g_cmds.cpp
+++ b/src/game/g_cmds.cpp
@@ -4457,14 +4457,18 @@ void Cmd_Goto_f(gentity_t *ent) {
     }
   }
 
+  Utilities::copyStanceFromClient(ent, other);
   VectorCopy(other->client->ps.origin, ent->client->ps.origin);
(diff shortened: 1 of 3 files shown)


src/game/g_cmds.cpp
```diff
@@ -4540,14 +4544,18 @@ void Cmd_Call_f(gentity_t *ent) {
     }
   }
 
+  Utilities::copyStanceFromClient(other, ent);
```
ovska, line 4547: These bits of `call` and `goto` code look identical, could maybe move to a common function such as `void MovePlayerToOther(gentity_t *player, gentity_t *other)`

ovska (approved):

PR #1243, opened 2024-03-05
Fix OB detector not getting correct shared value on demo playback

Aciz:
`cg_pmove.shared` is always 0 on demo playback, whereas `cgs.shared` has the correct value.


Aciz:
refs c8d4ee3845f426a58b1a2968d7ab2367bde48bec

PR #1246, opened 2024-03-05
Fix grenade weapon entity pickup issues

Aciz:
* Fix pickup models for both grenades
* Fix early exit on `Pickup_Weapon` that broke item pickups for other weapons

Still need to fix spinning flags for both models (axis doesn't rotate at all, allies doesn't rotate correctly around the model origin), probably need to add stand models for them.

Allies grenade pickup text is currently "Pineapple", might want to change this too (requires editing `animations/scripts/human_base.script` which has the grenade labeled as `pineapple` but unsure of potential consequences of this)

refs #1195


ovska (approved):

Aciz:
New pickup models added so the grenades don't sink halfway into the ground, and allies grenade pickup text fixed with a dirty hack for now, will investigate the script editing and it's potential consequences later.

image

Aciz:
Seeing spinning is broken for some other `weapon_` entities too, should probably look into that in more depth later (#1247).

PR #1261, opened 2024-03-15
Cleanup dynamic shader cleanup calls

Aciz:
Sending nullptr for both shadername and shadertext will clear up all dynamic shaders, there's no need to keep track of them on cgame side.

refs #159, #800


Aciz:
Actually this did clean them all up correctly, but it was just unnecessary, there is no need to keep track of them on cgame side as a single call will handle cleanup of everything.

PR #1272, opened 2024-03-24
feature: Simple snaphud

numbfingies:
gets rid of small snap zones

### DIFF
--- a/src/cgame/etj_accel_color.cpp
+++ b/src/cgame/etj_accel_color.cpp
@@ -94,7 +94,8 @@ void AccelColor::calcAccelColor(pmove_t *pm, vec3_t &accel, vec4_t &outColor) {
         frameAccel *
         static_cast<float>(std::sin(DEG2RAD(accelAngleAlt + altOptAngle))));
 
-    if (pm->groundPlane) {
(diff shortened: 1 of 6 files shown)


Aciz:
Please rebase.

Aciz:
I must say that I also really do not see the point of this, why would you want to do this? What benefit does it bring?

numbfingies:
> I must say that I also really do not see the point of this, why would you want to do this? What benefit does it bring?

This helps to see zones by their impact on advancing 'forwards'. Essentialy, this removes small snap zones and simplifies HUD.
This feature needs more work. Converting to draft

Aciz:
> > I must say that I also really do not see the point of this, why would you want to do this? What benefit does it bring?
> 
> This helps to see zones by their impact on advancing 'forwards'. Essentialy, this removes small snap zones and simplifies HUD.

But it does still show small zones though, when highlighting is enabled?

image

I just don't see how this benefits anyone, if you are hiding the zones, you are essentially giving incorrect information to the user, as they would instinctively aim to the end of a zone, but now this is actually incorrect as you're aiming at an "invisible" zone that is not optimal for accel.

PR #1277, opened 2024-04-02
Cgaz + Snaphud refactor

numbfingies:
optimized code to reduce redundant calculations.

### DIFF
--- a/src/cgame/etj_cgaz.cpp
+++ b/src/cgame/etj_cgaz.cpp
@@ -429,8 +429,6 @@ float CGaz::getOptAngle(const playerState_t &ps, pmove_t *pm, bool alternate) {
     wishspeed = static_cast<float>(ps.speed) * ps.sprintSpeedScale;
   }
 
-  UpdateDraw(wishspeed, &ps, pm);
-
   // no meaningful value if speed lower than ground speed or no user input
   if (state.vf < state.wishspeed || (cmd.forwardmove == 0 && cmd.rightmove == 0)) {
     return 0;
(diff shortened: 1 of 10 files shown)


numbfingies:
etj_CGaz1DrawSnapZone causes crash. need to investigate

numbfingies:
ok, fixed

src/cgame/etj_cgaz.cpp
```diff
@@ -443,43 +441,32 @@ float CGaz::getOptAngle(const playerState_t &ps, pmove_t *pm, bool alternate) {
...
+    opt = velAngle + perAngle - accelAngle;
+  } else if (cmd.rightmove > 0) {
+    // fullbeat / halfbeat / invert (holding +moveright)
+    opt = velAngle - perAngle - accelAngle;
+  } else if (cmd.forwardmove != 0) {
+    // nobeat
+    opt = velAngle + perAngle;
   }
 
   // return minimum angle for which you still gain the highest accel
```
Aciz, line 472: `opt` is uninitialized if none of these conditions is true.

src/cgame/etj_cgaz.cpp
```diff
@@ -429,8 +429,6 @@ float CGaz::getOptAngle(const playerState_t &ps, pmove_t *pm, bool alternate) {
     wishspeed = static_cast<float>(ps.speed) * ps.sprintSpeedScale;
   }
 
-  UpdateDraw(wishspeed, &ps, pm);
-
   // no meaningful value if speed lower than ground speed or no user input
   if (state.vf < state.wishspeed || (cmd.forwardmove == 0 && cmd.rightmove == 0)) {
```
Aciz, line 433: Removing this, while good since this unnecessarily updates the entire CGaz state, now breaks anything that uses `getOptAngle` as it relies on `state.vf/wishspeed` which are not calculated.

Aciz (changes requested):
Please also see build logs and clean the warnings.

src/cgame/etj_cgaz.cpp
numbfingies: Fixed now. added this back and moved updatesnap out of updatedraw method. :)

Aciz (approved):

PR #1300, opened 2024-04-10
Move angle vectors from pml_t to pmove_t

Aciz:
This improves accuracy of wishvel for cgaz/snaphud for spectators/demo playback, as pml_t can be wildly inaccurate on interpolation.


Aciz:
This also helps with flickering of `etj_CGaz1DrawSnapZone`.

PR #1308, opened 2024-04-12
Disable fast floating point math

Aciz:
Disable fast floating point math from release mode compile options, and instead just disable user-facing math trapping and floating point exception handling. This means we also now support NaN.

This fixes a flickering issue with `etj_CGaz1DrawSnapZone` which was caused by a fp math error in `AnglesToRange`. Need to test this a bit more still as there might be some other affected things now that we allow NaN.


Aciz:
Side note: this specific fast math bug with `etj_CGaz1DrawSnapZone` seems to not affect MSVC builds, but it does affect GCC builds (haven't tested Clang)

PR #1311, opened 2024-04-15
Fix corona frustum culling with etj_realFov 1 + optimize drawing

Aciz:
Coronas were incorrectly culled from view too early when etj_realFov 1 was used. Also some slight optimizations to the code in `CG_Corona`.

### DIFF
--- a/src/cgame/cg_ents.cpp
+++ b/src/cgame/cg_ents.cpp
@@ -1466,7 +1466,10 @@ void CG_Corona(centity_t *cent) {
 
     // use the dot to at least do trivial removal of those behind you.
     dot = DotProduct(dir, cg.refdef_current->viewaxis[0]);
-    if (dot >= cg.refdef_current->fov_x) {
+    const auto fov = static_cast<float>(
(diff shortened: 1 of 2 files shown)


src/cgame/cg_ents.cpp
```diff
@@ -1439,66 +1439,55 @@ CG_Corona
...
-  if (dot >= -0.6) // assumes ~90 deg fov	(SA) changed value to
-                   // 0.6 (screen corner at 90 fov)
-  {
-    behind = qtrue; // use the dot to at least do trivial
-                    // removal of those behind you.
-  }
-  // yeah, I could calc side planes to clip against, but would that be
-  // worth it? (much better than dumb dot>= thing?)
+    // performance variable cg_coronafardist will keep down super long traces
+    if (dist > static_cast<float>(cg_coronafardist.integer)) {
```
ovska, line 1463: I'd either use `static_cast<int>(dist)` or use `cg_coronafardist.value`. I doubt the <1.0 makes a difference.

ovska (approved):

PR #1313, opened 2024-04-16
Add warning messages about missing misc_gamemodels/server models

Aciz:
This mainly aids mappers in making sure that every required model that needs to displayed on runtime is included in the map pk3.

image


ovska (approved):
nice change

PR #1320, opened 2024-04-18
Send teaminfo updates to both teams

### LINKED ISSUE #1305: Update fireteam status for all teams
Our fireteams are team-independent, but fireteam status updates are only processed on a per-team basis, meaning Allied players on fireteam don't get health/class etc. updates from Axis players on the same fireteam.

Aciz:
Since our fireteams are shared between all teams, this ensures that players on Axis get the health info of players on Allies and vice versa. ~~This was actually already the case if a player had swapped teams once before, but this makes sure that it's always communicated, even if no team swaps ever occur.~~ No it was not, it would just not update the health at all and keep the last known value.

fixes #1305

### DIFF
--- a/src/game/g_team.cpp
+++ b/src/game/g_team.cpp
@@ -936,9 +936,9 @@ void TeamplayInfoMessage(team_t team) {
   bufferedData = team == TEAM_AXIS ? level.tinfoAxis : level.tinfoAllies;
 
   tinfo = va("tinfo %i%s", cnt, string);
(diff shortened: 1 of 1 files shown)


Aciz:
This needs some logic to split the info into multiple packets potentially as it might be too big to send in a single packet.

Aciz:
Actually, each client's info is sent as a single packet, so the only worry here would be to overflow the server command buffer, which is unlikely to happen just due to this as this adds only a few commands most of the time.

PR #1324, opened 2024-04-19
add configurable portal size to func_portaltarget

### LINKED ISSUE #1319: Add ability to scale portal size that's fired on func_portaltarget
`portal_scale` key for `func_portaltarget` - allows scaling the size of the portal fired on the surface, e.g. `portal_scale 0.5` -> half size of normal portal. With `-1` value, portal could scale to be the exact size of the surface.

ovska:
closes #1319


Aciz (approved):

Aciz (commented):
Other than maybe capping the max portal size, lgtm :+1:

Issue #894, opened 2022-12-29
Issues related to client and server side fps


Aciz:
There are several things in the game code that are dependant on fps, both on client (`com_maxfps`) and server (`sv_fps`). The client side issues are of course currently affecting everyone, but these issues aren't (in the context of ETJump) that severe, however we should fix them regardless. The bigger problems come from server side fps dependencies. The biggest issue of them all is mapscripts - the `wait` key is tied to `sv_fps`. This is the main blockade of us running servers on higher than `sv_fps 20`.

The issue with mapscripts is as said, `wait` being tied to `sv_fps 20`. This cannot really be fixed per se, however we should implement a system that compensates for it. For example, right now with `sv_fps 20`, if you specify `wait 101` in mapscript, in reality this wait is going to be `wait 150`. This is because the server simply updates 20 times per second, meaning __50ms__ frametimes. So in such scenario, `wait 101` effectively turns into `wait 150`, because on the first frame there will be __51ms__ remaining on wait, second frame __1ms__, 3rd frame wait is over -> action is executed, however this is delayed by __49ms__ from the specififed time in the mapscript. In contrast to say running a server at `sv_fps 125`, we're getting __8ms__ frametimes on server. This means that eventually during that wait period, we'll land at a scenario where we have waited for __96ms__ (12 frames on `sv_fps 125`) and the next server frame will be at __104ms__ -> wait is over, action gets executed, but this time it will be __46ms__ earlier than on `sv_fps 20`.

To compensate for this, we should treat all (old) mapscripts as if the server is running at `sv_fps 20`. This means checking the server framerate in the `wait` code, and adjusting the real wait time so that it roughly aligns how things work at `sv_fps 20`, since that's what they were designed for. Assuming we're aiming for `sv_fps 125` servers, we cannot perfectly match things 1:1, for example for __50ms__ wait, the closest real wait we can do with `sv_fps 125` is either __48ms__ or __56ms__ (we should probably always round up). This difference will be so small though that it should not affect gameplay. For future maps, we could add a new implementation of `wait` key, or a script action that will tell the mod what fps the script should run at. This way new maps can take advantage of more precise `wait` times, while maintaining backwards compatibility with old maps. As a gradual effort, we can even start fixing up some old mapscripts to work properly with high fps.

A lot of work has already been done regarding fps dependency fixes in ET: Legacy, so we can farily easily port those fixes into ETJump. From the top of my head, the following things need fixing:

__Client side issues - `com_maxfps` related__

[...]

PR #1327, opened 2024-04-20
Fix server-side fps dependencies

Aciz:
Fixes to various things affected by increasing `sv_fps > 20`.

* `wait` key in mapscripts is emulating `sv_fps 20` behavior
* Disguise stealing is normalized to `sv_fps 20`
* Flamethrower flamechunks movement per frame normalized to `sv_fps 20`
* `WolfRevivePushEnt` push frequence normalized to `sv_fps 20`

There are still a few smaller things that are affected by high `sv_fps`, but these are the most important ones for our potential use case.

refs #894

### DIFF
--- a/src/game/g_script_actions.cpp
+++ b/src/game/g_script_actions.cpp
@@ -1534,6 +1534,14 @@ qboolean G_ScriptAction_Wait(gentity_t *ent, char *params) {
     }
     max = Q_atoi(token);
 
+    // get as close as possible to sv_fps 20 intervals for compatibility
+    if (sv_fps.integer > 20) {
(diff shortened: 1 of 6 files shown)


src/game/g_script_actions.cpp
```diff
@@ -1534,6 +1534,14 @@ qboolean G_ScriptAction_Wait(gentity_t *ent, char *params) {
     }
     max = Q_atoi(token);
 
+    // get as close as possible to sv_fps 20 intervals for compatibility
+    if (sv_fps.integer > 20) {
+      min = min + DEFAULT_SV_FRAMETIME - (min & DEFAULT_SV_FRAMETIME) -
```
ovska: Should probably be `min % DEFAULT_SV_FRAMETIME`, same below

ovska (approved):

PR #1331, opened 2024-04-22
Improve lagometer

Aciz:
* Display snapshot delta values instead of 999 ping in demo playback
* Add client/server snapshot rate display to lagometer

imageimageimage

### DIFF
--- a/src/cgame/cg_draw.cpp
+++ b/src/cgame/cg_draw.cpp
@@ -651,6 +651,29 @@ LAGOMETER
 ===============================================================================
 */
 
+namespace ETJump {
+static int getDemoSnapshotInterval() {
+  int interval;
+
+  // cgs.sv_fps is present
+  static bool svFpsAvailable = demoCompatibility->isCompatible({3, 3, 0});
+  // CS_SYSTEMINFO contains sv_fps key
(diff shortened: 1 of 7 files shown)


src/cgame/cg_draw.cpp
```diff
@@ -890,6 +985,51 @@ static void CG_DrawLagometer() {
...
+  vec4_t textColor = {0.625f, 0.625f, 0.6f, 1.0f};
+  float fps;
+
+  if (cg.demoPlayback) {
+    static bool svFpsAvailable =
+        ETJump::demoCompatibility->isCompatible({3, 3, 0});
+    static bool csAvailable =
+        ETJump::demoCompatibility->isCompatible({3, 2, 0});
+
+    if (svFpsAvailable) {
```
ovska, line 998: Could this whole piece of code be moved to democompat, e.g. `ETJump::demoCompatibility->getSvFps()`

ovska (approved):

src/cgame/cg_draw.cpp
Aciz: Done, moved it to `etj_utilities.cpp` instead as it's likely going to be used in the future a lot with pmove related drawing stuff, to determine if we can safely use pmove command time for timing intervals.

PR #1332, opened 2024-04-23
Add "trigger_teleport_client" for client side predicted teleports

Aciz:
~~This comes with few caveats but works for most use cases for `trigger_teleporter`. Multi-target teleporters are excluded (can't guarantee same randomly picked target on client/server) as well as spawnflags 4/8 (relative angle teleporters) as they get all sorts of messed up when client calculates the relative angles on same values multiple times per one server activation.~~

Adds `trigger_teleport_client` entity to enable client side predicted teleport events. Supports same features as regular `target/trigger_teleport` with few exceptions:
* No multi-target teleporters
* No spawnflags 4/8 (might enable in the future if I manage to make them work with prediction)


Aciz:
As a side note, this also fixes prediction for `trigger_push` triggers with non-axial planes (client was previously getting an axial bbox based off mins/maxs instead of correct trigger dimensions).

Aciz:
After further testing and considerations, enabling prediction for existing teleporters contains too many pitfalls with badly built maps, and since not all of the teleporter functions work correctly with prediction, I decided it makes more sense to create a new entity that is predicted, and only supports the things that work correctly.

ovska (approved):

ovska (approved):

PR #1333, opened 2024-04-25
Allow toggling player collision between fireteam members

Aciz:
Allows players to enable collision between fireteam members when `g_ghostPlayers` is set to `1`. This effectively reworks the way ghostplayers is handled in the mod entirely: players are no longer nonsolid with `g_ghostPlayers 1` by default, but instead we add each player to ignored entities for collision checking before Pmove is handled, if the player is not member of our fireteam, or if fireteam collision is disabled.

This also fixes a bug where noclipping players weren't ignored in `StuckInClient`, which caused them to push other players around when noclipping, if `g_ghostPlayers` was `0`.

* `fireteam rules noghost <1|on>/<0/off>` to enable/disable
  * fireteam menu entry is also available
* projectiles and bullets now collide with solid players (wasn't previously the case with `g_ghostPlayers 0`)
* worldspawn key `noftnoghost` which allows disabling fireteam noghost in a map
* added `spawnflag 256` to `target/trigger_starttimer` to allow turning on/off ghosting during a timerun
  * if fireteam noghost is enabled, any timerun lacking this spawnflag will not start
  * if a fireteam member is currently timerunning on a run without this flag, fireteam noghost cannot be turned on


Aciz:
Projectile and bullet collision is now enabled for solid players (no damage though). This was previously not the case when `g_ghostPlayers` was disabled, but considering it wasn't really ever used for much, I don't see problem in enabling this, and it just makes more sense anyway to not be able to shoot through solid players.

Aciz:
Timeruns are now handled, and `noftnoghost` ws key was added. This should be ready now.

Issue #349, opened 2018-06-06
Enhance chat


isRyven:
Chat requires some serious ux updates, so we should simply rewrite it instead.

#### Chat Output
- [ ] Allow in place dynamic changes for chat width and height, through current characters per line and lines counts adjustments. Currently chat width doesn't change dynamically.
- [ ]  Add distraction free modes:  
   1. only allow mentions to be shown in chat, all messages still gonna be available in console
   2. make all chat text gray colored, but mentions  
       Allow to change filtered messages color. This could also be done by making messages transparent, then we should provide a way to adjust transparency.
- [ ] Add chat mode icons(global, team, fireteam)
- [ ] Remove coordinates, or add option to toggle them   
       Coordinates are currently set from qagame along with message, so If we leave them, we should set them directly from cgame.
- [ ] Allow players to select default colors for their chat messages
- [ ] Indicate voice chats with analogues mode icons
- [ ] Indicate private messages with respective icon
- [ ] Allow popup and chat merge, to form a single window for messages 
- [x] Allow chat text size change (through scaling)
- [ ] Allow character width adjustments (through scaling)
- [ ] Allow adjusting line height
- [x] Chat history, that will remain after map change
- [ ] Allow changing chat background color
- [x] Make chat always fade to fully transparent on the right side and [fix that stupid blue line](https://i.imgur.com/TYWpNbj.png)
- [ ] Allow customization of chat fade out time

#### Chat input (Optional)
- [ ] Move textbox from ui to cgame
- [ ] Make textbox window attached to chat window  
   Allow ways to switch between old and new position.  
   The new way would of course require to redesign the window.  
- [ ] Allow switching between modes using key shortcut or tabs
- [ ] Add private message mode
- [ ] Autocompletion for mention and private names
- [ ] Allow changing `messageMode` in textbox with kb shortcuts (`CTRL + LEFT/RIGHTARROW`)

_Anyone can add more tasks by editing the original post_

PR #1335, opened 2024-04-28
Add chat replay functionality

Aciz:
Replays up to 10 last global chat messages to client when they connect to a server, or perform a `vid_restart`.

Few limitations:
* Chat timestamps are lost as they are generated on demand client side, and I'd rather not transmit them and start to perform timezone conversions
* Chat team flag might be wrong, as it's parsed on demand on client side, so if a message was sent as Allies, and the replay is performed after the player has swapped to Axis, it will show the Axis flag in chat instead

refs #349


Aciz:
Extended functionality to preserve chat messages during map changes. Upon shutdown, chat messages are now logged to `chatreplay.json` on server, and restored upon map loading, so they can be replayed back to clients.

PR #1341, opened 2024-05-01
Remove gametype related code

Aciz:
Removes all references to gametype-specific things in the mod code (with few exceptions).

Removed following stuff:
* Removed all gametypes from the `gameType_t` enum except for `GT_WOLF` -> renamed to `ETJUMP_GAMETYPE`
* Removed all server/client/ui cvars and variables that are only relevant to specific gametypes
* Removed all functions/function code paths that are only used in specific gametypes
* Removed all macros that are relevant to specific gametypes
* Removed `campaign` and `listcampaigns` commands
* Removed all references to single player and savegame support

Stuff that was still kept:
* `g_gametype` cvar is still present for server browser functionality
* `MAX_GAMETYPES` is still `16` for server browser functionality
* Host Game menu still contains some LMS/Stopwatch specific cvars, only removed the LMS subwindow. Gametype can also still be changed, but the map feeder only functions on Objective mode


Aciz:
Should be all done now, updated the original PR message to detail what was removed.

PR #1380, opened 2024-05-27
Fix iterator invalidation in event loop removal

### LINKED ISSUE #947: Quitting the game during autodemo stop delay causes a segfault
Happens in debug, on release the client seemingly exits normally.

suburbski:
Fixes #947


Aciz:
Could you rebase this so the checks aren't all fucked up, since the CI name has changed so it's expecting checks that are no longer available.

PR #1381, opened 2024-05-28
Distinguish between client and server say

### LINKED ISSUE #1234: Highlight text doesn't work for server say
image

suburbski:
Fixes #1234

### DIFF
--- a/src/cgame/cg_servercmds.cpp
+++ b/src/cgame/cg_servercmds.cpp
@@ -724,23 +724,25 @@ static void CG_ConfigStringModified(void) {
   }
 }
 
+constexpr int REPLAY_MSG = 1 << 0;
+constexpr int SERVER_MSG = 1 << 1;
+
 /*
 =======================
 CG_AddToTeamChat
 
 =======================
 */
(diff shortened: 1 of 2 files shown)


src/cgame/cg_servercmds.cpp
```diff
@@ -2377,8 +2380,8 @@ static void CG_ServerCommand(void) {
 
     CG_RemoveChatEscapeChar(text);
     CG_FixLinesEndingWithCaret(text, MAX_SAY_TEXT);
-    s = CG_AddChatModifications(text, Q_atoi(CG_Argv(2)), chatReplay);
-    CG_AddToTeamChat(s, Q_atoi(CG_Argv(2)), chatReplay);
+    s = CG_AddChatModifications(text, clientNum, chatReplay, serverSay);
+    CG_AddToTeamChat(s, clientNum, chatReplay, serverSay);
```
Aciz, line 2384: Maybe these functions could just take an arbitrary "message type" as an input instead of to two bools? e.g.
```cpp
REPLAY_MSG = 1 << 0;
SERVER_MSG = 1 << 1;

CG_AddToTeamChat(s, clientNum, (REPLAY_MSG | SERVER_MSG))
```

PR #1390, opened 2024-06-03
Add draw yaw

### LINKED ISSUE #873: Implement tj_drawYaw
TJMod's `tj_drawYaw`, shows little arrows left/right when moving mouse left/right.

suburbski:
Closes #873


src/cgame/etj_draw_yaw.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+  const playerState_t *ps = getValidPlayerState();
+  const int yawAngle = ANGLE2SHORT(ps->viewangles[YAW]);
+  const int diff = yawAngle - oldYawAngle;
+  constexpr int halfRange = 65536 / 2;
+  moveLeft = ((diff > 0 && diff <= halfRange) || diff < -halfRange);
+  moveRight = ((diff < 0 && diff >= -halfRange) || diff > halfRange);
+  oldYawAngle = yawAngle;
+
+  return true;
```
Aciz, line 62: Why not `return (moveLeft || moveRight)` here?

suburbski: there's an early out now

src/cgame/cg_local.h
```diff
@@ -2810,6 +2810,13 @@ extern vmCvar_t etj_ccMenu_width;
 extern vmCvar_t etj_ccMenu_browseWithOpen;
 extern vmCvar_t etj_ccMenu_showEmptyPages;
 
+extern vmCvar_t etj_drawYaw;
+extern vmCvar_t etj_drawYawX;
+extern vmCvar_t etj_drawYawY;
+extern vmCvar_t etj_drawYawColor;
+extern vmCvar_t etj_drawYawShadow;
+extern vmCvar_t etj_drawYawSize;
```
Aciz: This isn't the cvar naming convention we have, these should be
```
etj_drawYaw
etj_yawX
etj_yawY
etj_yawColor
etj_yawShadow
etj_yawSize
```

Aciz (changes requested):
I wonder if we also want something like `etj_yawWidth`, which would "push" the markers away from the drawing point, to the sides? that way you could for example clearly have the indicators on left and right side of your crosshair.

Aciz:
Also change base to `dev` pls

Aciz (approved):

PR #1400, opened 2024-07-26
Fix misc_beam collision

Aciz:
Send `mins/maxs` to client + fix the incorrect origin for the beam start, was using the origin of the `misc_beam` entity, when it should have used the origin of the entity marking the start of the beam. Fixes dodgy movement around Railgun cranes mostly.

This is not perfect as it only works for straight axial lines but I don't think it's possible to get an inline model registered for this so it cannot work correctly on arbitrary angles, it just gets a solid rectangle bbox always.

It also doesn't quite work perfectly in Railgun when the cranes start moving, but at least the collision box is now actually correct and somewhat accurate.


Aciz:
Seems to collision issue with Railgun cranes once they start moving is due to the fact that `ent->s.pos.trBase` doesn't get updated every frame, but only when the movement direction of the crane changes, need to figure out why that's the case.

PR #1422, opened 2024-08-28
Add colorpicker for UI

Aciz:
Color picker menu to set colors for cvars that take color values.

https://streamable.com/zkap73

* Opened via `Color picker...` entry in dropdown menus, present in every color cvar dropdown
* Sliders for RGB (0-255/0-1), HSV and alpha
  * Resulting cvar value is always normalized RGB
* Interactive HSV color picker which can be used with mouse to pick a color
  * `mouse2` adjusts only saturation
  * `mouse3` adjusts only value
* Preview boxes for old and new color
* `RESET` button resets the color picker to a state at which it was when opened
* Adjusts the cvar in real time, so you'll see results instantly

Things to potentially improve in the future:
* Selectively disable alpha slider for cvars that have a separate alpha cvar (currently is adjusts the preview color, which is not representative of the actual color getting applied)
* ~~Allow switching the RGB sliders to 0-1 range~~ done
* Hex color input/preview box
* `CTRL` + `ALT` key modifiers for saturation/value only adjustments in color picker, this requires some rewriting to the mouse capture functionality though as it currently only supports capturing a single key


Aciz:
Added a button to toggle normalized/full RGB sliders

https://streamable.com/6uuq7w

PR #1431, opened 2024-09-01
Display full maplist in map vote menu

Aciz:
Request a full list of maps from server on cgame init, and forward the list to UI for display on `Vote -> Map` menu. This ensures clients will always have full list of maps available in map vote menu, even if the server only communicates referenced pak list due to configstrings overflow, or if the client doesn't have the map(s) downloaded.


Aciz:
Map lists are now also sorted alphabetically.

PR #1442, opened 2024-09-15
Refactor mapvote details menu

Aciz:
* Draw the map details with an actual menu item using ownerdraw
* Details panel now updates as you select a new map in the list, instead of requiring a new click on details button
* Fix levelshot aspect ratio, was 1:1 when it should be 4:3
* Add text scroll to map briefing if text doesn't fit, also used in Host Game menu

https://streamable.com/dutgcu


Aciz:
The text now fades out when going out of bounds, instead of harshly disappearing https://streamable.com/qq9ynu

PR #1456, opened 2024-09-22
Add savepos system

### LINKED ISSUE #1402: Add savepos

Aciz:
* `savepos <optional name> <optional flags>`
* `loadpos <optional name>`
* `listsavepos` - lists all valid savepos files
* `readsavepos` - reads savepos files (for refreshing in runtime)

Flags:

* `1` - Don't store speed (speed is cleared)
* `2` - Don't store pitch angle (reset to 0)

Stores players position, angles, velocity, stance and timerun state into a file, which can then be restored using `loadpos`. Savepos is implemented fully in cgame side, meaning it can be used to save a valid state from demo playback (timerun state only works for demos recorded in ETJump 3.0.0 or newer). `loadpos` requires cheats to be enabled.

This probably requires a few more tweaks until it's fully finished, ~~for example I'd like to add JSON parsing error prints for cgame, currently they are only for qagame sided logger class~~ done.

fixes #1402


Aciz:
I think this is done now, one thing that does not work that I'm not sure I'll bother is `connectNotify`, meaning if you load a timerun state with `loadpos`, and another client connects and starts spectating you, they won't see the timer. It works fine if you're already on the server, just not if the `loadpos` command happened while you were not connected.

Aciz:
Checkpoints are currently borked for timeruns, because we don't have the information on exactly which checkpoints are hit, only the number of checkpoints. We can't get this information currently, so we need to modify the checkpoint command here and tack the checkpoint index at the end of the checkpoint server command, so client knows *which* checkpoint was hit:

https://github.com/etjump/etjump/blob/ddad5a0937167bdf6c7f7263317c00c00c8adc2c/src/game/etj_timerun_v2.cpp#L368-L379

The information then needs to be parsed in the server commands, and stored here:

https://github.com/etjump/etjump/blob/ddad5a0937167bdf6c7f7263317c00c00c8adc2c/src/cgame/etj_timerun.h#L51-L69

Unfortunately this means that it won't be possible to restore timerun state from demos recorded prior to 3.3.0, but this is required to make them work correctly.

This should be fine, and doesn't require any backwards compatibility layer for older demos for just pure playback, since the extra argument can be added at the end of the command, so we just need to check if it exists here at all:

https://github.com/etjump/etjump/blob/ddad5a0937167bdf6c7f7263317c00c00c8adc2c/src/game/etj_timerun_shared.cpp#L153-L190

PR #1488, opened 2024-10-15
Add g_chatReplayMaxMessageAge

Aciz:
Can be set to > 0 to make chat replay messages expire. Any chat message older than current time - `g_chatReplayMaxMessageAge <minutes>` will be excluded from chat replay. The chat is still included in the replay file, it just won't be sent to clients. Off by default (messages don't expire).

refs #1335


Aciz:
Side note: we might be able to restore message timestamps now, but I will take a look at that in the future.

PR #1498, opened 2024-10-19
Fix etj_smoothAngles with snipers

Aciz:
Sniper recoil is handled entirely in Pmove, so we must disable smooth angles while scoped with a sniper.

refs #1495


Aciz:
Closing in favor of #1499

PR #1508, opened 2024-10-22
Add timestamps to chat replay messages

Aciz:
Adds relative timestamps to chat replays (1h ago, 30min ago, 10s ago etc.). Goes up to 12h, older messages are displayed as `12h+ ago`.

refs #1335


Aciz:
Rather than absolute timestamps, opted to making the timestamps relative, as they are a bit easier to read and to quickly grasp how old a message is.

image

PR #1511, opened 2024-10-24
Fix trigger_teleport_client for spectators

Aciz:
~~This breaks prediction as noclipped players aren't predicted, and spectators don't get linked so setting origin breaks.~~ `ET_TELEPORT_TRIGGER_CLIENT` wasn't handled in `G_TouchTriggers`.

refs #1332


Aciz:
Prediction now works correctly for spectators, the entity was just ignored in `G_TouchTriggers` for specs. Noclip is still not predicted, which I'm not sure if it's even fixable.

PR #1531, opened 2024-11-11
Add support for engine extensions + flash window on chat mention/pm

Aciz:
Initial implementation of engine extension system. Wired to each module, currently only used in cgame for flashing the game window on chat mentions and incoming private messages.


Aciz:
Window flash on chat mentions/PMs is now behind `etj_highlight & 2`, instead of being implicitly enabled whenever chat highlighting is enabled.

PR #1577, opened 2024-12-06
Increase max brush entity limit

Aciz:
The engine supports 512, and since bmodels aren't actually stored/communicated in CS even though the limit was checked against a CS limit, there should be no issues in increasing this to the same value as engine uses.

~~I will still need to test how this behaves if someone tries to for example add 500 brush entities with a `model2` key set, because registering models is still limited by CS.~~ Should be okay, see comment below.

### DIFF
--- a/src/cgame/cg_local.h
+++ b/src/cgame/cg_local.h
@@ -2013,8 +2013,8 @@ typedef struct {
   sfxHandle_t gameSounds[MAX_SOUNDS];
 
   int numInlineModels;
-  qhandle_t inlineDrawModel[MAX_MODELS];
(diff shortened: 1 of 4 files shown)


Aciz:
> I will still need to test how this behaves if someone tries to for example add 500 brush entities with a `model2` key set, because registering models is still limited by CS.

This should actually be fine, the server never let's you register more than `MAX_MODELS` anyway, so this would just error with configstrings overflow.

https://github.com/etjump/etjump/blob/85e707a6fcf5f5bc9ff285efc66c7286735a7b55/src/game/g_utils.cpp#L81-L115

PR #1617, opened 2025-02-18
Add etj_onDemoPlaybackStart cvar

Aciz:
Value gets executed on first frame of demo playback (and unfortunately also if using `vid_restart` during playback, but there's no way around that).

This very much compliments #1616 for automatic, mass video recording, but it's entirely detached from that so I made it a separate PR.

I will pray to the git gods that git can figure out a simple line addition to a plaintext file but I don't have much faith so this will likely conflict and needs to be rebased.


Aciz:
Renamed the cvar to `etj_onDemoPlaybackStart` to be more explicit that this is about demo playback, not demo recording.

ovska (approved):

Issue #1588, opened 2024-12-26
Add entity for setting fireteam rules


Aciz:
`target_ft_setrules` - entity to set fireteam rules for the activators fireteam. This is very convenient for maps that are made to be played with collision or nosave, and essentially a requirement to make proper teamjump maps after #1357 is merged, to prevent cheating.

Aciz:
Implemented in #1625

PR #1625, opened 2025-02-23
Add target_ft_setrules entity + supplementary ws keys

Aciz:
Allows setting fireteam rules for the activcators fireteam. Bypasses `noftnoghost` and new `noftsavelimit` & `noftteamjumpmode` worldspawn keys, so mappers have precise control over when these rules are set. Spawnflag 1 makes it so that only the leader of the fireteam is able to activate the entity.

Supported keys:
* `savelimit (-1 - 100 or reset)` - sets savelimit
* `noghost 0/1` - disable/enable fireteam noghost
* `teamjumpmode 0/1` - disable/enable teamjump mode
* `leader_only_message` - message to print to activator if `spawnflag 1` is set and the activator is not the fireteam leader, `%s` will be substituted with the fireteam leaders name

flxes #1588

### DIFF
--- a/src/game/g_spawn.cpp
+++ b/src/game/g_spawn.cpp
@@ -17,6 +17,7 @@
 #include "etj_missilepad.h"
 #include "etj_target_init.h"
 #include "etj_trigger_teleport_client.h"
+#include "etj_target_ft_setrules.h"
 
 qboolean G_SpawnStringExt(const char *key, const char *defaultString,
                           char **out, const char *file, int line) {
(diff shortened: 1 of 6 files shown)


src/game/g_spawn.cpp
```diff
@@ -176,6 +177,11 @@ field_t fields[] = {
 
     {"cursorhint", FOFS(s.dmgFlags), F_CURSORHINT},
 
+    {"savelimit", FOFS(ftSavelimit), F_LSTRING},
+    {"noghost", FOFS(damage), F_INT},
+    {"teamjumpmode", FOFS(health), F_INT},
+    {"leader_only_message", FOFS(ftLeaderOnlyMsg), F_LSTRING},
```
Aciz, line 183: `savelimit` and `leader_only_message` don't actually work here, `F_LSTRING` is made to work on char pointers, not arrays.

PR #1638, opened 2025-03-02
Automatically center mapscript brush origins

Aciz:
Automatically set the origin of mapscript-spawned brushes to be at the center of the brush, and adjust mins/maxs accordingly. This makes sure brushes get properly assigned to correct PVS when linked, but still lets people create brushes in mapscripts by using the simple method of setting mins to "0 0 0" and maxs to the brush size.


Aciz:
Actually upon further investigation, this doesn't matter in the slightest and is entirely useless.

PR #1695, opened 2025-06-04
Add etj_useExecQuiet for map/team autoexecs

Aciz:
When enabled, map and team-specific autoexecs are executed using `execq`, to omit the print about executing a config file from the console output.

Hidden from the menus on 2.60b as it does not support `execq` (which is why we can't just enable this implicitly)


Aciz:
Changed this to be a bitflag instead, so this can be done individually per autoexec "type".

PR #1703, opened 2025-07-26
Add countdown command for fireteams

Aciz:
A simple `countdown` command for fireteams that prints a countdown to fireteam chat. This can be used to easily coordinate timings on teamjumps, without having to be in voice chat or write scripts yourself.

If given an argument, the countdown will last for the specified amount of seconds. If no arguments are given, the default time will be taken from `etj_fireteamCountdownLength` cvar value. The countdown is capped to 1-10s. If the cvar is <= 0, a default timer of 3s will be used.

The countdown can also be started via the fireteam menu.

image


Aciz:
Moved the command to be handled as generic `fireteam` command, so the command is now `fireteam countdown` rather than just `countdown` to clarify it's a fireteam command.

PR #1793, opened 2025-10-10
Add vote UI spectator count

### LINKED ISSUE #872: Vote UI should separate spectator votes from player votes
This makes it clearer why vote might fail even in situations where it seemingly should pass.

Example scenario:
* 4 players, 4 specs
* vote is called by a player, 3 other players don't vote at all, all specs vote yes
* vote tally is 5-0 yes-no visually, but vote fails because 3 players didn't vote = counts as 3 no votes

gytispieze:
Resolves #872

Added spectator vote counting for regular and RTV vote UI. Follows the format `TOTAL_VOTES (SPEC_VOTES)`

image

image

### DIFF
--- a/src/cgame/cg_draw.cpp
+++ b/src/cgame/cg_draw.cpp
@@ -2165,7 +2165,6 @@ CG_DrawVote
 =================
 */
 static void CG_DrawVote() {
-  char str1[32], str2[32];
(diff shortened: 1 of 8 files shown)


Aciz (approved):
Fixed re-vote stuff + added demo backwards compatibility stuff, should be good now.

PR #1809, opened 2026-01-03
Improve perceived smoothness of mouse cursor

Aciz:
This makes mouse movement work on real screen pixels rather than virtual grid pixels. Previously, the visual "steps" that the mouse would move in UI was restricted to the resolution of the virtual grid, which made the mouse feel choppy and inaccurate, especially on smaller movements. The visual position is now tracked on real screen pixels, which are then converted to virtual grid coordinates for actual UI interaction.

In the future, we could also consider tracking the real screen coordinates for UI interaction, since this would allow sliders to have more granularity on the drag operations.

### DIFF
--- a/src/cgame/cg_draw.cpp
+++ b/src/cgame/cg_draw.cpp
@@ -4414,16 +4414,9 @@ static void CG_Draw2D() {
     CG_DrawCHS();
   } else {
     if (cgs.eventHandling != CGAME_EVENT_NONE) {
-      //			qboolean old =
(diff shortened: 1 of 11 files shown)


Aciz:
Should note that this is a somewhat breaking change, as it overall lowers the base UI sensitivity, so players will likely need to tweak `etj_menuSensitivity` to their liking again. Imo this is such a minor inconvenience that it did not justify the added the complexity of modifying the sensitivity calculations to produce same results as before.

PR #1823, opened 2026-01-27
Simplify clientside player collision resolving

Aciz:
Rather than filtering this out everywhere where it's relevant, just figure out which players are solid when building the solid entity list, so we don't need to take extra care everywhere when tracing against the world, as it just works automatically if the client doesn't think the entities are solid to begin with.

refs #1333

### DIFF
--- a/src/cgame/cg_draw.cpp
+++ b/src/cgame/cg_draw.cpp
@@ -1915,26 +1915,6 @@ static float CG_ScanForCrosshairEntity(float *zChange, qboolean *hitClient) {
   return dist;
 }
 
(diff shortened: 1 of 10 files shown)


Aciz:
Actually this will not work correctly, because all traces are now down purely from the observing players perspective. For example, if players A and B are in a fireteam with collision enabled, and player C shoots player A, player B will think that player C hit player A, because from the perspective of player B, player A is solid, and therefore should block things like bullets. Currently, this works because we're filtering players individually, and always from the shooters perspective, rather than our own.

PR #1864, opened 2026-03-06
Add high-res, anti-aliased drawing for CGaz2

### LINKED ISSUE #1861: Add anti-aliasing for cgaz2 lines

Aciz:
Adds an alternative, high resolution drawing for CGaz2, using Xiaolin Wu's line drawing algorithm. This is an anti-aliased line drawing algorithm, which is performed on real pixel coordinates rather than the virtual grid for accuracy. This new drawing is enabled by default, and can be turned off with the `etj_CGaz2HighRes` cvar. Accommodating this is also `etj_CGaz2Thickness` cvar, which controls the line thickness. This is only applied to the high-res drawing, and defaults to 2, which is visually close to the original drawing. The valid range is 0.5 - 5.0.

This drawing is relatively expensive compared to the old method (about 4-5x slower), but the absolute cost is still quite reasonable (~65000ns on a Ryzen 7 3700X). The more concerning part is the amount of draw calls that this generates, which might cause issues on 2.60b, but oh well, you can't have it all.

Old:
old

New:
new

fixes #1861


Aciz:
> The more concerning part is the amount of draw calls that this generates, which might cause issues on 2.60b, but oh well, you can't have it all.

This actually already is an issue on ETe/ETL with > 1080p resolutions, there's simply too many draw calls that the UI starts to break. Will have to revisit this and see if anything could be done, but unfortunately this might have to be reverted.

PR #1873, opened 2026-04-14
Add target_worldspawn_modifications entity - selectively ignore ws keys

### LINKED ISSUE #1573: Allow nosave zones to only be enabled during timeruns
Could be implemented with a worldspawn key, e.g. `nosave_timerun`. This would make it simpler to make sure timeruns don't have unwanted exploits with save/load in them, while keeping the map more relaxed while casually playing.

Aciz:
This entity let's mappers apply some worldspawn keys selectively to players instead of as a global setting. Keys can be made to be ignored for each team, and while timerunning/not timerunning. The supported keys are following:

* `nogod`
* `nogoto`
* `nojumpdelay`
* `nonoclip`
* `nosave`
* `nooverbounce`
* `noprone`
* `nowallbug`
* `overbounce_players`
* `portalgun_spawn`

The team(s)/timerun state is selected with spawnflags:

* __1__ - ignore for Axis
* __2__ - ignore for Allies
* __4__ - ignore for Spectators
* __8__ - ignore when timerunning
* __16__ - ignore when not timerunning

At least one team must be selected for each entity. Both timerun states cannot be set on the same entity - if you want to ignore timerun state completely, don't set either of the keys.

When a key is ignored, all checks related to the key's behavior are ignored. This means that for example for keys which have some surface/contentparm associated with them (e.g. `nosave`), the behavior is *NOT* reversed, instead any restrictions will simply be ignored, and the ignored team(s) can save anywhere, regardless of save zones. Similarly keys like `nojumpdelay` and `nooverbounce` completely ignore the surfaceparms instead of flipping - ignored `nojumpdelay` means jump delay is enabled everywhere, regardless of surfaceparms.

You may have multiple `target_worldspawn_modifications` entities in a map, and their rules will be applied individually. Do not try to apply rules for the same key for a team, in multiple entities - this will cause unpredictable behavior.

This entity does not count towards the map's entity limit.

Radiant entity definition:
```xml
<point name="target_worldspawn_modifications" color="1.0 0.7 0.3" box="-8 -8 -8 8 8 8">
Selectively ignore worldspawn keys.
-------- KEYS --------
<boolean key="nogod" name="No God">Apply properties set by this entity to "nogod" key behavior.</boolean>
<boolean key="nogoto" name="No Goto">Apply properties set by this entity to "nogoto" key behavior.</boolean>
<boolean key="nojumpdelay" name="No Jump Delay">Apply properties set by this entity to "nojumpdelay" key behavior.</boolean>
<boolean key="nonoclip" name="No Noclip">Apply properties set by this entity to "nonoclip" key behavior.</boolean>
<boolean key="nosave" name="No Save">Apply properties set by this entity to "nosave" key behavior.</boolean>
<boolean key="nooverbounce" name="No Overbounce">Apply properties set by this entity to "nooverbounce" key behavior.</boolean>
<boolean key="noprone" name="No Prone">Apply properties set by this entity to "noprone" key behavior.</boolean>
<boolean key="nowallbug" name="No Wallbug">Apply properties set by this entity to "nowallbug" key behavior.</boolean>
<boolean key="overbounce_players" name="Overbounces on players">Apply properties set by this entity to "overbounce_players" key behavior.</boolean>
<boolean key="portalgun_spawn" name="Portalgun spawn">Apply properties set by this entity to "portalgun_spawn" key behavior.</boolean>
-------- SPAWNFLAGS --------
<flag key="IGNORE_AXIS" name="Ignore Axis" bit="0">The selected key(s) will be ignored by Axis players.</flag>
<flag key="IGNORE_ALLIES" name="Ignore Allies" bit="1">The selected key(s) will be ignored by Allied players.</flag>
<flag key="IGNORE_SPECTATORS" name="Ignore Spectators" bit="2">The selected key(s) will be ignored by Spectators.</flag>
<flag key="TIMERUN_ONLY" name="Timerun only" bit="3">The restrictions set by the selected key(s) will be ignored only when timerunning.</flag>
<flag key="NO_TIMERUN" name="No timerun" bit="4">The restrictions set by the selected key(s) will be ignored only when not timerunning.</flag>
    -------- NOTES --------
This entity does not need to target/be targeted to anything, similar to how "script_multiplayer" works.
You can have multiple "target_worldspawn_modifications" entities in the map, and their properties will be applied individually.
This entity does not count towards the map entity limit.

Multiple teams may be selected to ignore a key via same entity. Each entity must target at least one team.
Do not try to ignore the same key for a team in two different entities. This will cause unpredictable behavior.

"TIMERUN_ONLY" and "NO_TIMERUN" cannot be selected simultaneously. If you want to ignore timerun status completely, leave both unchecked.
</point>
```

fixes #1573


Aciz:
Superseded by #2015

PR #1892, opened 2026-05-07
Add PAIN spawnflag for func_static_client, don't fire targets when toggled

Aciz:
Similar to `func_static`, activates targets when the entity takes damage. Cannot be killed by dealing too much damage, the entity health is restored upon taking damage. Predictably only works when the entity is visible/on, as it is ignored in traces when it's invisible/off. The damage must be dealt by a client, damage dealt by world (e.g. `shooter_rocket`) is ignored.

Made this `spawnflag 2` and pushed up the other spawnflags, to match regular `func_static`.

refs #1824


Aciz:
Targets are no longer fired when the entity is toggled - this made no sense as any targets you might want to fire when you toggle the entity can be fired via the entity that toggles this instead. Now `spawnflag 2` can fire a unique target, that doesn't get fired when the entity is toggled.

PR #1906, opened 2026-05-17
Fix CGaz1DrawSnapZone

suburbski:
Effectively extends CGaz1 min angle to the end of the current snapzone, and hides all other CGaz1 info when active.


Aciz (approved):
please add a short description on what this actually fixes for searchability, otherwise good

PR #1952, opened 2026-07-22
Add snaphud cropping

Aciz:
* `etj_snapHUDCrop 0-2` - allows users to crop the screen coordinates where snaphud is being drawn. The default behavior is to crop everything from the center of the screen, against the current strafe direction (so if player is strafing right, everything left of the crosshair will not draw, and vice versa). If the value is 1, the crop is only performed when strafing, value 2 crops always.
* `etj_snapHUDCropOffsets <center> <edge>` - offsets the cropping points from the center/edge of the screen against the current strafe direction, respectively. When offset is applied to the center, more of snaphud will draw on the cropped side of the screen center. When offset is applied to the edge, the snaphud drawing cuts off before the edge of the screen. The values can be combined as such that the entire snaphud can be made "cropped and centered", e.g. by setting `<center>` to 200, and `<edge>` to 320 - 200 = 120. Valid range for both values is **0.0 - 320.0**

Some examples, strafe direction is left on all pictures:

Default crop:
2026-07-22-132617-long

`etj_snapHUDCropOffsets 100 0`
2026-07-22-132627-long

`etj_snapHUDCropOffsets 100 100`
2026-07-22-132632-long

`etj_snapHUDCropOffsets 200 120`, "centered and cropped"
2026-07-22-132645-long


Aciz:
Changed the behavior of `etj_snapHUDCrop` such that when set to **1**, the cropping is only applied when the player is strafing. Value **2** crops always, regardless of input status. This probably desired, as it's a bit weird that half of the snaphud is gone when the player isn't strafing, since default input is applied to snaphud drawing.

PR #1960, opened 2026-07-30
Add in-game crosshair editor

kareta0:
Adds an editor for building your own crosshair in game, instead of picking one of the built in ones. It sits behind a button on the existing crosshair page, Settings > Basic HUD > Crosshair > Open Editor.

### What it does

A crosshair is up to 8 elements stacked on top of each other. Per element you get a shape (cross, lines, diagonal cross, dot, circle, square, triangle, T, V, two lines), separate X and Y size and gap so anything can be stretched, thickness, rotation, offset, and primary, secondary and outline colours with their own alphas.

Elements are stored one per cvar in `etj_crosshairElement1..8` as a compact `key=value;key=value` string. That way a single element always fits in MAX_CVAR_VALUE_STRING, it goes into the config like everything else, and you can still set one from the console if you want. `etj_customCrosshair` switches the whole thing on, `etj_customCrosshairScale` scales it.

Two part shapes use the secondary colour for their second half, and filled shapes use it as the fill under a primary border, so it behaves like `cg_crosshairColorAlt` does on the built in ones.

### Where the code sits

The shape maths is in `src/cgame` (`etj_crosshair_definition`, `etj_crosshair_renderer`) and ui builds those two files as well, the same way it already builds `etj_color_parser`. ui needs the exact same drawing for the editor preview and the two modules have their own 2d drawing apis. Each one passes its own `fillRect` and `drawLine` in through `CrosshairPainter`, so the preview is literally the same code that draws in game and cannot drift from it. cgame had no circle helper so there are scanline filled ellipses and triangles in there too. Those are capped at 480 rows, otherwise a crosshair sized to cover the screen would cost hundreds of draw calls a frame.

`Crosshair::render` takes the custom path only when `etj_customCrosshair` is set, the existing crosshairs are untouched. The element cvars are parsed through `CvarUpdateHandler` when they change rather than every frame, so colours are resolved once.

### The editor itself

.menu binds a widget to a cvar by name at parse time, so there is no way to point a slider at one field inside a packed element string. Instead there is a working cvar per field (`etj_ce*`) for the widgets to bind to, and the editor syncs those against the selected slot in both directions each frame. Side effect worth having: console edits and preset loads show up in the widgets straight away. It also meant the existing colour picker and combos could be reused as they are.

Every slider row has a small edit box on the same cvar next to it. A slider can only resolve `(max - min) / SLIDER_WIDTH` per pixel and SLIDER_WIDTH is fixed at 96, so on a range wide enough to be useful you can never land on the value you actually want. Dragging gets you close, typing gets you exact.

The syncing runs off the preview ownerdraw, which only paints while an editor page is open, which is also the only time any of it needs to happen.

### Presets

Plain text files, one serialized element per line, so they stay readable and easy to pass around. Saving and loading are on the editor page and the presets page, and the list is a normal listbox on a new feeder, click to pick and double click to load.

They live in `profiles/<profile>/crosshairs/`. That is not arbitrary: `FS_AllowDeletion` only lets the engine delete under `save/`, `profiles/` and `demos/`, so a preset kept anywhere else could be written and read but never deleted.

### Notes for review

- One new ownerdraw (`UI_CROSSHAIR_EDITOR_PREVIEW`) and one new feeder (`FEEDER_CROSSHAIR_PRESETS`).
- Three new menu files, plus a button added to `etjump_settings_hud1_crosshair.menu`. `NUM_MAIN_BTNS` and the main tab bar are untouched, I had it as its own top level tab at first and it was not worth narrowing every button in every settings menu for one feature.
- `ui_shared.cpp` is unchanged.
- Definitions written by an older build still load, unknown keys are ignored on parse.
- Not tested on linux or 32 bit, only x64 windows so far.


assets/ui/menudef_ext.h
```diff
@@ -93,6 +93,56 @@
...
+    backcolor       .3 .3 .3 .4 \
+    style           WINDOW_STYLE_FILLED \
+    border          WINDOW_BORDER_FULL \
+    bordercolor     .1 .1 .1 .5 \
+    cvar            cvarname \
+    maxChars        8 \
+    maxPaintChars   7 \
+    visible         1 \
+    tooltip         "Click and type an exact value" \
+  }
```
Aciz: This should be in `menumacros_ext.h` rather than here

assets/ui/etjump_settings_hud1_crosshair.menu
```diff
@@ -10,6 +10,12 @@
 #define NUM_SETTINGS 10 // required for bottom-up settings drawing
 #define SIDEBAR_BTN_Y (SIDEBAR_Y + (SIDEBAR_H * 0.5) - (((NUM_SUBMENUS * SIDEBAR_BTN_H) + (MAIN_ELEMENT_MARGIN * (NUM_SUBMENUS - 1))) * 0.5))
 
+// button into the crosshair editor, sits in the gap under the picker
+#define EDITOR_BTN_W 150
+#define EDITOR_BTN_H 18
+#define EDITOR_BTN_X (SETTINGS_SUBW_X + ((SETTINGS_SUBW_W - EDITOR_BTN_W) * 0.5))
+#define EDITOR_BTN_Y (SETTINGS_ITEM_Y + 96 + 10) // 96: crosshair picker height
```
Aciz: Put this `96` size as a constant in `menudef_ext.h` and use that here, this is already getting defined twice in both `etjump_settings_crosshair_colors.menu` and `etjump_settings_crosshair_editor.menu`

src/cgame/etj_crosshair.h
```diff
@@ -40,6 +41,14 @@ class Crosshair : public IRenderable {
   void adjustPosition();
   static bool canSkipDraw();
 
+  // reads etj_crosshairElement1..8 into definition. only runs when one of those
+  // cvars changes so we resolve colors once instead of every frame.
+  void parseDefinition();
+
+  // every slot cvar, so the listener setup and the parse loop share one list
+  static std::array<const vmCvar_t *, CrosshairLimits::maxElements>
+  elementCvars();
```
Aciz: This could just be a public `static constexpr std::array<...>`, no? Or is there a reason it's a function call? I don't see any at least

src/cgame/etj_crosshair_renderer.cpp
```diff
@@ -0,0 +1,474 @@
...
+// from the center.
+//
+// two-part shapes (cross, V, two lines...) draw one part in the primary color
+// and the other in the secondary, and filled shapes use the secondary as the
+// fill, same as the built-in crosshairs do with cg_crosshairColorAlt.
+
+// diagonal arms get projected onto both axes so a diagonal cross ends up the
+// same visual size as a straight one with the same numbers
+constexpr float diagonalFactor = 0.70710678f; // sqrt(0.5)
+constexpr float pi = 3.14159265358979323846f;
```
Aciz: `M_PI` in `q_shared.h`

src/cgame/etj_crosshair_renderer.cpp
```diff
@@ -0,0 +1,474 @@
...
+  float thickness;
+  float rotation;
+};
+
+Point rotatePoint(const Point p, const Point pivot, const float degrees) {
+  if (degrees == 0.0f) {
+    return p;
+  }
+
+  const auto radians = degrees * (pi / 180.0f);
```
Aciz: `DEG2RAD` in `q_shared.h`

Aciz (commented):
Haven't fully looked through this yet and not tested very thoroughly, but a couple of comments:

* The element editor UI really needs a "reset element" button
* The sliders in the element editor don't correctly respond to clicks on the slider, and write completely bogus values

Overall this seems like a great addition, although it does require some more polish. Given this provides everything that the ETJump vector crosshairs do and more, down the line I think it would make sense to deprecate those in favor of this, and provide presets with the mod to match the current vector crosshairs. But that's a breaking change, so it won't be necessary until next major release.

assets/ui/menudef_ext.h
kareta0: Moved to `menumacros_ext.h`.

assets/ui/etjump_settings_hud1_crosshair.menu
kareta0: Done, it's `CROSSHAIR_PREVIEW_SIZE` now.

src/cgame/etj_crosshair.h
kareta0: No reason, changed it to a public `static constexpr std::array`.

src/cgame/etj_crosshair_renderer.cpp
kareta0: Fixed.

src/cgame/etj_crosshair_renderer.cpp
kareta0: Fixed.

kareta0:
Both fixed. Reset button sits next to the preset save, it keeps the shape since `Empty` already clears a slot.

The slider issue was in `Item_Slider_HandleKey`: clicks didn't clamp or snap the way dragging does. It affects every slider, not just the editor, so it's in its own commit in case you want it separate.

kareta0:
Also fixed two things I ran into while testing: filled shapes ignored rotation, and their fill alpha blended against the outline behind them instead of the world.

PR #1976, opened 2026-08-19
Glob translation units in CMakeLists.txt files

Aciz:
All modules + tests now just glob translation units, so there's no need to explicitly add them anymore, apart from cross-module TUs. CMake automatically picks up reconfiguration with `CONFIGURE_DEPENDS`.

Also cleaned duplicate listing in tests CMakeLists.txt


Aciz:
microslop can't handle this

PR #1990, opened 2026-08-26
Improve default settings for clients

Aciz:
A lot of this is obviously opinionated, but I believe these are all strictly improvements over the current out-of-box experience.

* `cg_drawCrosshar` **1 -> 5**
* `cg_crosshairSize` **48 -> 40**
* `cg_crosshairPulse` **1 -> 0**
* all text shadow cvars enabled
* `etj_realfov` **0 -> 1**
* `etj_stretchCgaz` **1 -> 0**
* bunch of tweaks to CGaz1 & snaphud defaults
* `etj_drawspeed` **1 -> 0** (drawspeed2 is much more useful)
* `cg_drawRoundTimer` properly defaults to **1** (was already the case in cgame, but UI overwrote the default)
* `etj_hideFadeRange` **200 -> 128**
* `etj_drawOB` **0 -> 1**
* OB & slick detector default positions changed - both are now below crosshair such that they don`t overlap with CGaz & snaphud (most notably CGaz 2), slick detector to the left, OB detector to the right
* `etj_popupTime` **1000 -> 0** (delay between popup messages)
* `cg_drawSmallPopupIcons` **0 -> 1**
* `cg_bloodFlash` **1.0 -> 0.0**
* `etj_autoDemo` **0 -> 1**


Aciz:
`cg_drawCrosshair 5` does not unfortunately work since the engine forces default value to be 1, but I'm leaving it here anyway, in case some source port stops doing this stupidness one day.

PR #2010, opened 2026-09-18
Build system updates and improvements

Aciz:
* Split the entire build & package processes into a reusable `build/package.yml` workflows. Both accept ref as an input so it works for both dev builds and tagged releases. Upload is handled separately from the package step (was previously combined).
* Automatic dev builds (latest tag) no longer automatically trigger on push to `dev`. Instead, they run daily on schedule, or on manual trigger, via the new `snapshot.yml` workflow. If there are no new commits since the latest build, the build is simply skipped. This is handled in a new "Check" step that runs before the build.
* macOS builds now specify `Ninja` as generator explicitly. By default, CMake uses XCode as a generator, which is insanely slow - `Ninja` generator achieves about 7 times faster builds. Since macOS builds no longer take close to 30mins, The split build + manual `lipo` binary combination is no longer done, and everything runs in a single job now. This frees up macOS runners on periods of high activity, since GH Actions free limits only allow 5 concurrent macOS runners.
* Updated `paths-ignore` to `ci.yml` to use YAML anchors to avoid maintaining 2 lists, plus updated some of the ignored files.
* Replaced `czietz/action-automatic-releases` with `softprops/action-gh-release`, as the former is archived and unmaintained. The same release action is now used for both `latest` build and tagged releases.

With this, I am also turning off merge queue from the repo. It made sense when the dev builds were triggered on every commit, to reduce the churn of automatic updates on the dev server, but because we now run on schedule, there's not much point in having it enabled.


Aciz:
Tested all sorts of things with this on my own fork and everything seems to work, merging this to master and to dev from there, so that the branches will have synced build systems.

